PR #25, opened 2026-04-03
feat: add compiler validation layer and generalized testing

tiero:
Introduces a two-pass validation pipeline and comprehensive test coverage
as the quality gate for the v0.1.0-alpha release.

## Semantic validator (src/validator/mod.rs)
- Pre-compile AST pass: rejects duplicate function names, duplicate
  constructor params, contracts with no non-internal functions, missing
  exit timelock when server key is used, and zero exit timelock values.
- Post-compile output pass: asserts structural invariants on emitted JSON
  (non-empty ASM, witnessSchema, both server+exit variants present per
  function) to catch compiler bugs before output reaches callers.
- 12 unit tests covering all validation rules inline.

## Pipeline integration
- compiler/mod.rs: runs validate_ast() after parse (fatal errors halt
  compilation); runs validate_output() after code generation (issues
  surface as warnings, non-blocking).
- lib.rs + main.rs: declare validator module so both library and binary
  crate roots resolve crate::validator.

## Bug fix
- compiler/mod.rs:2080: replaced .unwrap() on Option<&String> with an
  explicit if-let guard, making the intent clear and eliminating the
  implicit panic path.

## Generalized tests
- tests/compilation_roundtrip_test.rs: 17 tests — one per example .ark
  file plus a sweep that compiles the entire examples/ directory and
  asserts output invariants (non-empty name, functions, ASM,
  witnessSchema, both variants per function).
- tests/validation_error_test.rs: 13 tests — covers empty source,
  syntax errors, zero timelock, duplicate names, no functions, internal-
  only contract, and validates error message quality.

Total test count: 139 (was 98).

https://claude.ai/code/session_012kj2GGGAsZoFK8m1GUinMA

## Release Notes

* **New Features**
  * Enhanced compiler validation with comprehensive semantic checks for contract structure, function definitions, and assembly output
  * Improved error reporting with detailed diagnostic messages for validation failures
  * New warnings for potential issues including missing guard statements and timelock configurations

* **Tests**
  * Added comprehensive test coverage for validation, type-checking, and assembly structure verification

PR #28, opened 2026-05-08
Add StabilityVault contract and comprehensive PRD

tiero:
## Summary

Adds the **StabilityVault** product: a USD-denominated claim backed by BTC collateral, enabling Arkade Wallet users to hold a stable dollar balance on Bitcoin without exchange or issuer risk. This PR ships the production contract, its supporting `StabilityOffer` and `PriceBeacon` contracts, the PRD, and integration tests, and propagates the new Arkade contract conventions (canonical `options { server = server; exit = exit; }`, `<SERVER_KEY>` auto-injection, 330-sat Taproot dust) across the rest of the codebase.

## Stability product

**`examples/stability/stability_vault.ark`** — the live USD position.

- `transfer(seekerSig, newSeekerPk)` — seeker assigns the full position to a new owner (recursive covenant).
- `split(seekerSig, amountUSD, newSeekerPk)` — seeker divides the USD claim; collateral splits proportionally; both halves enforced ≥ 330 sats (Taproot dust).
- `seekerRedeem(seekerSig)` — seeker settles to BTC at the live beacon price.
- `providerExit(providerSig)` — provider settles to BTC at the live beacon price.

Settlement math (sats / cents, no floating point):

```
seekerBase     = targetUSD * 1e8 / currentPrice
fundingAccrued = fundingSatPerBlock * (tx.time - openHeight)
seekerRaw      = seekerBase + fundingAccrued
seekerPayout   = clamp(seekerRaw, 0, totalCollateral)
providerPayout = totalCollateral - seekerPayout
```

Three branches per settlement function: provider takes everything when funding has eaten the seeker's claim (`seekerRaw <= 0`); seeker takes everything when undercollateralised (`seekerRaw >= totalCollateral`); otherwise a proportional split, with the provider output dropped when its payout is at or below the 330-sat dust floor.

**Settlement model: first come, first served.** `seekerRedeem` and `providerExit` are economically symmetric — same beacon read, same payout formula. There is no counterparty-challenge window and no settlement delay; whoever broadcasts first locks in the live oracle price for both sides. The `exit` timelock is purely a fallback for when the Arkade Operator is offline.

**`examples/stability/stability_offer.ark`** — provider's standing offer.

- `take(userBTC, seekerPk)` — non-interactive: any swap service or seeker can claim the offer at the live beacon price, producing a fresh `StabilityVault`.
- Beacon output index is conditional on whether a remaining-offer output is created (`output[1]` if fully consumed, `output[2]` otherwise).
- `withdraw(providerSig)` — provider reclaims unused collateral at any time.

**`examples/stability/price_beacon.ark`** — dual-asset oracle.

- `ticker` asset quantity = BTC/USD price in cents.
- `clock` asset quantity = block height of the last update.
- `update` enforces monotone advancement of the clock — no back-dating.
- `passthrough` is enforced on every read, and the script is checked
  (`tx.inputs[N].scriptPubKey == new PriceBeacon(ticker, clock, oraclePk, exit)`) so consumers can authenticate the feed.
- Freshness: `tx.time - tx.inputs[N].assets.lookup(clock) <= 144` (≈ 24 hours).
- `migrate(oracleSig, newOraclePk)` rotates the oracle key while preserving both assets.

## Convention enforcement (codebase-wide)

These rules are now codified in `CLAUDE.md` and applied to every example contract:

- `options { server = server; exit = exit; }` — never bind `server` to anything other than `server`, never hardcode the exit timelock as a literal.
- `int exit` (and `int renew` where relevant) live in the constructor so the playground can parameterise them.
- The Arkade Operator key is auto-injected as `<SERVER_KEY>` in ASM — it never appears as a constructor parameter.
- Taproot dust threshold is **330 sats** everywhere.
- Always "Arkade", never "ARK" or "ASP".

These convention changes touch every `.ark` and `.json` artifact under `examples/`, plus the inline test fixtures in `tests/threshold_oracle_test.rs` and `tests/beacon_test.rs`.

## Compiler changes

`exit` and `renew` in the options block now accept either an integer literal (`exit = 144`) or an identifier referencing a constructor parameter (`exit = exit`). The compiler emits the literal directly or a `<exit>` placeholder respectively. This is the language-level support that makes the new convention possible.

- `src/models/mod.rs`: `exit_timelock` and `renewal_timelock` changed from `Option<u64>` to `Option<String>`.
- `src/parser/mod.rs`: stores the raw token regardless of whether it parses as `u64`.
- `src/compiler/mod.rs`: emits the literal when the string parses as `u64`, otherwise emits `<identifier>`.

## Bug fixes from review

| ID | Issue | Fix |
|---|---|---|
| C1 | `seekerRedeem`/`providerExit` had no branch for `seekerRaw <= 0`, leaving funds locked when funding overran the seeker's claim | Added explicit `if (seekerRaw <= 0)` branch directing all collateral to the provider |
| C2 | `tx.time` was treated as unix seconds in the staleness check but as block height elsewhere | Standardised on block height throughout: `clock` asset stores block height, staleness = 144 blocks, `beaconHeight` naming |
| C3 | `take()` checked the beacon at a fixed output index even when no remaining-offer output existed | Beacon index now switches between `output[1]` and `output[2]` based on whether the remaining offer is created |
| H1 | `split()` did not enforce that the residual half was above dust | Added `require(collateralB >= 330, "remainder too small")` |
| H2 | Settlement branches only verified `ticker` survival, not `clock` | All beacon-passthrough requires now check both assets |
| M1 | Mixed dust thresholds (547 in some places, 330 in others) | Standardised to 330 everywhere |

## Documentation

- `docs/stability-vault-prd.md` — complete PRD: problem statement, economic model, user flows, contract architecture, settlement math, oracle design, and risk disclosures. Updated to match the implemented "first-come, first-served" settlement model and the dual-asset oracle.
- `README.md` — updated to use the canonical contract conventions (`server = server`, parameterised `exit`/`renew`) and to refer to "Arkade Operator" rather than "Ark operator".

## Files changed

- `examples/stability/{price_beacon,stability_offer,stability_vault}.ark`
- `examples/stability/{price_beacon,stability_offer,stability_vault}.json`
- `examples/*.{ark,json}` — convention propagation
- `examples/stable_position.ark` — deleted (superseded by `StabilityVault`)
- `examples/beacon.ark` — deleted (loop-unrolling fixture, superseded by `tests/beacon_test.rs::BEACON_LOOP_CODE`)
- `src/{models,parser,compiler}/mod.rs` — `exit`/`renew` identifier support
- `tests/beacon_test.rs` — dual-asset PriceBeacon test fixture + production tests
- `tests/threshold_oracle_test.rs` — convention update (server key removed from constructor)
- `playground/{main.js,generate_contracts.sh}` — recursive scan of `examples/`, `stability_vault` wired into examples list
- `docs/stability-vault-prd.md`, `README.md`, `CLAUDE.md` — documentation and convention rules

https://claude.ai/code/session_01WUTCmexF3MZg2hvEapmfCz

## Summary by CodeRabbit

* **New Features**
  * Stability Vault ecosystem for USD-denominated BTC positions with oracle pricing and settlement
  * Contract options now support parameterized `exit` and `renew` timelocks instead of hardcoded values

* **Documentation**
  * Language guide updated with standardized contract patterns and witness specifications
  * Product requirements document for Stability Vault

* **Refactor**
  * Contract witness schemas enhanced with explicit Schnorr signature encodings
  * Server key injection standardized across all contracts
  * Legacy oracle contract examples consolidated into production-ready implementations


ghost (changes requested):
# 🔍 Arkana Code Review — StabilityVault Contract Suite

**Verdict: Request changes.** This is protocol-critical code (settlement math, collateral splitting, oracle reads). Three critical bugs must be fixed before merge. Several high/medium issues also flagged.

---

## 🔴 CRITICAL

### C1. Missing `seekerRaw >= 0` clamp — negative funding can lock the UTXO

**`stability_vault.ark:153-165` (seekerRedeem) and `:212-224` (providerExit)**

The PRD specifies `seekerPayout = clamp(seekerRaw, 0, totalCollateral)`. The contract implements the upper bound (`if seekerRaw >= totalCollateral`) but **never clamps to zero**.

When `fundingSatPerBlock < 0` (seeker pays provider — explicitly supported per PRD §4) and enough blocks elapse, `seekerRaw` goes negative. In the `else` branch:

```
require(tx.outputs[0].value >= seekerRaw, "seeker underpaid");  // always true for negative seekerRaw
int providerPayout = totalCollateral - seekerRaw;               // > totalCollateral!
require(tx.outputs[1].value >= providerPayout, ...);            // impossible to satisfy
```

**Result:** The UTXO becomes unspendable through the normal settlement path. Provider and Seeker funds are permanently locked. This is a fund-loss bug.

**Fix:** Add a third branch:
```
if (seekerRaw <= 0) {
    // Provider takes everything, seeker gets nothing
    require(tx.outputs[0].value >= totalCollateral, "provider underpaid");
    require(tx.outputs[0].scriptPubKey == new SingleSig(providerPk), "output 0 not provider");
    // beacon at output[1]
} else if (seekerRaw >= totalCollateral) {
    // existing undercollateralised branch
} else {
    // existing normal branch
}
```

### C2. `tx.time` cannot be both block height and unix timestamp

**`stability_vault.ark:152-155` vs `:139-140`**

The contract mixes two incompatible time domains in the same transaction:

```ark
// Block-height arithmetic:
int blocksElapsed = tx.time - openHeight;     // openHeight is "block height at position open"
int funding = fundingSatPerBlock * blocksElapsed;

// Unix-timestamp arithmetic:
int beaconAge = tx.time - beaconTime;         // beaconTime is "unix seconds"
require(beaconAge <= 86400, "stale oracle");  // 86400 seconds = 24 hours
```

`nLockTime` (and therefore `tx.time`) is ONE value per transaction — either a block height (< 500000000) or a unix timestamp (≥ 500000000). You can't use it for both.

If `tx.time` is block height: `beaconAge = blockHeight - unixTimestamp` → garbage.
If `tx.time` is unix time: `blocksElapsed` is actually seconds, not blocks, and `fundingSatPerBlock` is misnamed (it's per-second). `openHeight` is also a misnomer.

**This needs a design decision:**
- Option A: Use unix timestamps throughout. Rename `openHeight` → `openTimestamp`, `fundingSatPerBlock` → `fundingSatPerSecond`, `blocksElapsed` → `secondsElapsed`.
- Option B: Use block heights throughout. Change the beacon `clock` to store block heights and adjust the staleness check (86400 blocks ≈ 60 days, not 24 hours — you'd want ~144 blocks for 24h).

Same issue exists in `stability_offer.ark:70-71`.

### C3. Beacon output index hardcoded to `output[2]` in `take()` — breaks when offer fully consumed

**`stability_offer.ark:95-101`**

```ark
if (remainingCapacity > 0) {
    require(tx.outputs[1].scriptPubKey == new StabilityOffer(...), ...);
}
// ALWAYS checks output[2], even when no output[1] offer exists:
require(tx.outputs[2].assets.lookup(ticker) >= entryPrice, "beacon must survive");
```

When the offer is fully consumed (`remainingCapacity == 0`), there is no output[1]. The beacon is at output[1], not output[2]. The unconditional output[2] check will fail — **a full-consumption take is impossible**.

**Fix:** Make the beacon index conditional:
```ark
if (remainingCapacity > 0) {
    // ... existing output[1] checks ...
    require(tx.outputs[2].assets.lookup(ticker) >= entryPrice, "beacon must survive");
    require(tx.outputs[2].assets.lookup(clock) >= beaconTime, "beacon clock must survive");
} else {
    require(tx.outputs[1].assets.lookup(ticker) >= entryPrice, "beacon must survive");
    require(tx.outputs[1].assets.lookup(clock) >= beaconTime, "beacon clock must survive");
}
```

---

## 🟠 HIGH

### H1. No `collateralB` dust check in `split()`

**`stability_vault.ark:100-102`**

```ark
require(collateralA >= 546, "split amount too small");
// No check on collateralB!
```

A seeker can split `amountUSD = targetUSD - 1` which leaves `collateralB` at potentially sub-dust levels. The remainder position becomes an unspendable UTXO.

**Fix:** Add `require(collateralB >= 546, "remainder too small");`

### H2. Settlement functions don't verify beacon clock asset on output

**`stability_vault.ark:166-170` (seekerRedeem), `:223-227` (providerExit)**

`transfer()` and `split()` check BOTH ticker and clock assets survive on output. But `seekerRedeem()` and `providerExit()` only check ticker:

```ark
require(tx.outputs[1].assets.lookup(ticker) >= currentPrice, "beacon must survive");
// Missing: require(tx.outputs[1].assets.lookup(clock) >= beaconTime, ...);
```

The beacon's own `passthrough()` script enforces both, so this is defense-in-depth. But if the beacon is consumed via a different leaf (or there's a beacon script bug), the clock asset could be drained. Given this is protocol-critical, add the check.

### H3. `import` resolution across directory boundary

**`stability_vault.ark:1-2`**

```ark
import "single_sig.ark";
import "price_beacon.ark";
```

`stability_vault.ark` is in `examples/stability/` but `single_sig.ark` is in `examples/`. If the compiler resolves imports relative to the importing file, `single_sig.ark` won't be found. Needs verification that the compiler's import resolver handles parent directories or has an include path.

---

## 🟡 MEDIUM

### M1. Dust threshold inconsistency

- `split()` uses `546` (P2PKH dust limit)
- `seekerRedeem()`/`providerExit()` use `330` (Taproot dust limit)
- PRD §7 says `547`

For Taproot outputs (which these are), 330 is correct. Pick one value and use it consistently. I'd suggest 330 everywhere since all outputs are P2TR.

### M2. `generate_contracts.sh` uses `basename` — latent name collision

**`playground/generate_contracts.sh:22`**

```js
const name = path.basename(full, '.ark');
```

If two `.ark` files in different subdirectories share the same basename, the second `export const` overwrites the first. No collision today, but this is fragile. Consider using the relative path as the key (e.g., `stability_price_beacon`).

### M3. Funding rate preserved identically on both split children

**`stability_vault.ark:106-120`**

`fundingSatPerBlock` is an absolute rate, not per-sat. After a split, both children accrue funding at the original rate. A $50 child and a $4,950 child both accrue the same absolute funding. This may be intentional (simpler), but it means the funding-to-USD ratio differs significantly between the two children. Worth documenting the design intent.

---

## ✅ What looks good

- Settlement math (`seekerBase = targetUSD * 1e8 / currentPrice`) correctly uses integer arithmetic with no floating point.
- Beacon authentication via script comparison (`tx.inputs[1].scriptPubKey == new PriceBeacon(...)`) is solid — prevents spoofed beacons.
- Monotonically non-decreasing timestamp enforcement in `update()` prevents back-dating.
- Recursive covenant structure (transfer/split produce new StabilityVault, settlement is terminal) is clean.
- The "no liquidation" design eliminates entire classes of MEV and race condition bugs.
- Test coverage for the new PriceBeacon dual-asset design is good (timestamp monotonicity, passthrough preservation).
- PRD is thorough and well-reasoned.

---

## ⚠️ PROTOCOL-CRITICAL FLAG

This PR introduces settlement math, oracle reads, and collateral handling. C1 (negative seekerRaw) is a direct fund-loss vector. C2 (time domain confusion) affects every funding calculation. **Human review required before merge.**

---

*🤖 Reviewed by Arkana*

ghost (changes requested):
# Arkana Code Review — arkade-os/compiler#28

**Verdict: Request changes.** One critical bug blocks all cooperative vault transactions. Several high-severity issues below.

---

## 🔴 CRITICAL — PriceBeacon passthrough() broken for vault transactions

**Files:** `examples/stability/price_beacon.ark` (passthrough function), `examples/stability/stability_vault.ark` (all functions)

The `passthrough()` function hardcodes `tx.inputs[0]` and `tx.outputs[0]`:

```ark
// price_beacon.ark — passthrough()
int currentPrice = tx.inputs[0].assets.lookup(ticker);   // ← hardcoded index 0
require(tx.outputs[0].scriptPubKey == new PriceBeacon(...));  // ← hardcoded index 0
require(tx.outputs[0].assets.lookup(ticker) >= currentPrice); // ← hardcoded index 0
```

But **every** StabilityVault function places the beacon at `input[1]`:

```ark
// stability_vault.ark — all functions
require(tx.inputs[1].scriptPubKey == new PriceBeacon(...), "wrong beacon");
int currentPrice = tx.inputs[1].assets.lookup(ticker);
```

**Impact:** The compiler generates the function body for the server (cooperative) variant. When the beacon is spent as input[1] in a vault transaction, its passthrough tapleaf evaluates `tx.inputs[0]` (the vault, not the beacon) and `tx.outputs[0]` (the vault output, not the beacon output). Both checks fail. **The cooperative path — the normal transaction flow — is broken on-chain.**

The exit variant is unaffected (N-of-N CHECKSIG fallback replaces the body), but requiring oracle signature + CSV timelock for every vault operation defeats the product design.

**The old `beacon.ark` got this right** — it used `tx.input.current.scriptPubKey` (position-independent) and `tx.assetGroups` loop (aggregated asset preservation). The new beacon regressed by switching to hardcoded indices.

**Fix:** Use `tx.input.current` for self-references. For output verification, either parameterize the output index or use `tx.assetGroups` to verify total asset preservation across all outputs, similar to the old beacon pattern.

---

## 🟠 HIGH — No test coverage for StabilityVault or StabilityOffer

**Files:** `tests/` directory

The PR adds 4 new tests for PriceBeacon but **zero** for StabilityVault (370 lines, 4 functions, settlement math) or StabilityOffer (139 lines, 2 functions). These are the most complex, money-handling contracts in the PR. At minimum:

- Compilation smoke test (parses without error)
- Structure test (correct number of functions × 2 variants)
- Settlement math: verify `seekerBase` / `funding` / `clamp` logic appears in ASM
- Split: verify two recursive outputs with proportional collateral
- Edge cases: zero funding, negative funding, undercollateralization branch

---

## 🟠 HIGH — Code duplication: seekerRedeem() and providerExit() are identical

**File:** `examples/stability/stability_vault.ark`

`seekerRedeem()` and `providerExit()` contain **exactly the same** settlement logic (~70 lines each) — beacon auth, staleness check, settlement math, 3-way branching (seekerRaw ≤ 0, seekerRaw ≥ totalCollateral, normal), dust handling. The only difference is which key signs.

This is a maintenance hazard. A bug fix applied to one but not the other creates a divergence in payout math — which is a funds-loss vector. Extract the shared settlement logic or add a comment block requiring synchronized edits.

---

## 🟡 MEDIUM — PRD contradicts code on dust threshold

**Files:** `docs/stability-vault-prd.md` (line ~327), `examples/stability/stability_vault.ark`

PRD says:
> Dust threshold: if `providerPayout < 547 sats`, no provider output is created.

Code uses 330 sats:
```ark
if (providerPayout > 330) {
```

The code is correct (330 is the Taproot dust limit). Update the PRD to match.

---

## 🟡 MEDIUM — PRD uses "ASP" terminology

**File:** `docs/stability-vault-prd.md`, Section 3 (Actors table)

The actors table defines **"ASP — Ark Service Provider"**. Per project rules in CLAUDE.md: *"NEVER use 'ASP' or 'ARK' — always 'Arkade'."* Replace with "Arkade Operator" throughout.

---

## 🟡 MEDIUM — Split rounding is undocumented and slightly asymmetric

**File:** `examples/stability/stability_vault.ark`, `split()` function

```ark
int collateralA = totalCollateral * amountUSD / targetUSD;  // truncates
int collateralB = totalCollateral - collateralA;            // gets remainder
```

Integer division truncation means `collateralB ≥ collateralA` even when `amountUSD = remainingUSD`. The remainder holder (original seeker) always gets ≥ their proportional share. Not exploitable (settlement is based on `targetUSD`, not raw collateral), but worth a comment explaining the rounding direction.

---

## 🟢 LOW — transfer() requires beacon validation

**File:** `examples/stability/stability_vault.ark`, `transfer()` function

`transfer()` reads and validates the beacon (freshness check, asset preservation), but it only changes `seekerPk` — no settlement math. This adds overhead (beacon must be included as input) and a liveness dependency (stale oracle blocks transfers).

If this is intentional (prevent transferring when oracle is stale), add a comment explaining why. If not, consider removing the beacon dependency from transfer.

---

## ℹ️ Notes

- **Cross-repo impact is low.** No external repo consumes the compiler JSON output directly. The `server = server` convention change is cosmetic (parser ignores the RHS). `witnessSchema` was already in the compiler before this PR.
- **Compiler change (parameterized exit timelock)** looks correct. The `u64` → `String` change in `src/models/mod.rs` and the `parse::<u64>()` guard in `src/compiler/mod.rs` handle both literal and identifier cases properly.
- **Potential merge conflict** with PR #25 which also modifies `src/compiler/mod.rs`.

---

⚠️ **Protocol-critical PR.** This touches settlement math, covenant recursion, and oracle verification — all protocol-level concerns. Even after fixing the above, this PR requires human review before merge.

ghost (approved):
# Arkana Code Review — arkade-os/compiler#28 (incremental, commits since last review)

**Reviewing:** `27659b82b0` (playground exit convention) and `9d89ae6b7e` (drop oraclePk from vault/offer)

**Verdict: Approve** — the architectural change is sound. No new bugs introduced.

---

## Analysis: Dropping `oraclePk` from StabilityVault / StabilityOffer

This is the significant change. Previously, consumers verified the beacon via scriptPubKey reconstruction:
```
require(tx.inputs[1].scriptPubKey == new PriceBeacon(ticker, clock, oraclePk, exit), "wrong beacon");
```

Now, consumers authenticate the beacon solely by its `ticker` and `clock` asset IDs via `tx.inputs[1].assets.lookup(ticker)`.

### Why this is safe

1. **Asset IDs are unforgeable.** In the Elements asset model, `bytes32` asset IDs are commitments to the genesis issuance transaction. No third party can mint assets with matching IDs.

2. **Beacon assets can't be stolen or split by third parties.** The PriceBeacon UTXO can only be spent via its script — cooperative path (requires server sig) or unilateral path (requires `oraclePk` sig + CSV). A malicious actor can't split the beacon's asset quantities into a fake UTXO.

3. **Oracle trust model is unchanged.** The oracle operator (holder of `oraclePk`) is already trusted to provide honest prices via `update()`. If they're malicious, they could corrupt prices regardless of whether consumers check the scriptPubKey.

4. **Key rotation becomes safe.** Removing `oraclePk` from consumer contracts means `PriceBeacon.migrate(newOraclePk)` works seamlessly — existing vaults don't need to be re-created after oracle key rotation. This is a real operational improvement.

### Files verified

| File | Change | Status |
|------|--------|--------|
| `examples/stability/stability_vault.ark` | Remove `oraclePk` param, drop `import "price_beacon.ark"`, remove 4× scriptPubKey checks | ✅ Correct — all `new StabilityVault(...)` references updated consistently |
| `examples/stability/stability_offer.ark` | Remove `oraclePk` param, remove scriptPubKey check, update `new StabilityVault(...)` and `new StabilityOffer(...)` calls | ✅ Correct |
| `examples/stability/price_beacon.ark` | Comment updates only | ✅ Accurate |
| `docs/stability-vault-prd.md` | Updated trust model documentation | ✅ Matches implementation |
| `examples/stability/stability_vault.json` | Regenerated artifact — `oraclePk` removed from constructorInputs, ASM updated | ✅ Consistent with .ark source |
| `examples/stability/stability_offer.json` | Regenerated artifact — same | ✅ Consistent |
| `playground/arkade-language.js` | `exit = 144` → `exit = exit` | ✅ Matches CLAUDE.md convention |

### Minor note (non-blocking)

The JSON artifacts still carry compiler warnings about implicit `uint64le`↔`scriptnum` conversions. These are pre-existing and not introduced by this diff, but worth tracking for a future cleanup pass.

---

⚠️ **Protocol-critical reminder:** This PR modifies VTXO settlement contracts and oracle authentication. Per Arkana protocol, **human review is still required** before merge even though the code is architecturally sound.

*🤖 Reviewed by Arkana — incremental review of commits `27659b82b0..9d89ae6b7e`*

ghost (changes requested):
# Arkana Code Review — arkade-os/compiler#28 (incremental, commit `43f0503`)

**Reviewing:** `43f0503` (docs: clarify oracle update cadence is sub-block on Arkade)

---

## 🔴 Documentation ↔ Contract Contradiction (docs/stability-vault-prd.md)

The new PRD text states:

> The `clock` field still records the block height at which the update was issued, so **multiple intra-block updates share that block height** — that is correct

But the `PriceBeacon.update()` contract (examples/stability/price_beacon.ark:44) enforces **strict** monotonicity:

```
require(newBlockHeight > currentHeight, "block height must advance");
```

This is `>`, not `>=`. Two updates at the same block height will fail the second time because `newBlockHeight == currentHeight` does not satisfy `>`.

**Either:**
1. The contract should use `>=` to allow same-block updates (but then back-dating within the same height is possible — the oracle could replay an older price at the same height), **or**
2. The PRD should not claim sub-block updates "share that block height." Instead, the oracle must use a synthetic monotonic counter (e.g., `block_height * 1000 + sub_index`) — but then the freshness check `tx.time - clock <= 144` breaks because `clock` would no longer be a real block height, and `tx.time - (height*1000+idx)` would be a large negative number.

**Recommendation:** If sub-block cadence is genuinely desired, introduce a separate sequence counter for intra-block ordering and keep `clock` as the actual block height with `>=`. The staleness check remains valid, and you get monotonic ordering. This requires a contract change, not just a docs change.

As written, the PRD makes a promise the contract cannot keep. This is dangerous for integrators who read the PRD and assume sub-block updates work.

---

Everything else in this commit (wording change from "on-chain UTXO" to "Arkade UTXO", actor table update) is fine.

⚠️ **Protocol-critical reminder:** This PR modifies VTXO settlement contracts and oracle semantics. Per Arkana protocol, **human review is still required** before merge.

*🤖 Reviewed by Arkana — incremental review of commit `9d89ae6b7e..43f0503d48`*

ghost (changes requested):
# Arkana Code Review — arkade-os/compiler#28 (incremental, commit `43f0503`)

**Reviewing:** `43f0503` (docs: clarify oracle update cadence is sub-block on Arkade)

---

## 🔴 Documentation ↔ Contract Contradiction (docs/stability-vault-prd.md)

The new PRD text states:

> The `clock` field still records the block height at which the update was issued, so **multiple intra-block updates share that block height** — that is correct

But the `PriceBeacon.update()` contract (examples/stability/price_beacon.ark:44) enforces **strict** monotonicity:

```
require(newBlockHeight > currentHeight, "block height must advance");
```

This is `>`, not `>=`. Two updates at the same block height will fail the second time because `newBlockHeight == currentHeight` does not satisfy `>`.

**Either:**
1. The contract should use `>=` to allow same-block updates (but then back-dating within the same height is possible — the oracle could replay an older price at the same height), **or**
2. The PRD should not claim sub-block updates "share that block height." Instead, the oracle must use a synthetic monotonic counter (e.g., `block_height * 1000 + sub_index`) — but then the freshness check `tx.time - clock <= 144` breaks because `clock` would no longer be a real block height, and `tx.time - (height*1000+idx)` would be a large negative number.

**Recommendation:** If sub-block cadence is genuinely desired, introduce a separate sequence counter for intra-block ordering and keep `clock` as the actual block height with `>=`. The staleness check remains valid, and you get monotonic ordering. This requires a contract change, not just a docs change.

As written, the PRD makes a promise the contract cannot keep. This is dangerous for integrators who read the PRD and assume sub-block updates work.

---

Everything else in this commit (wording change from "on-chain UTXO" to "Arkade UTXO", actor table update) is fine.

⚠️ **Protocol-critical reminder:** This PR modifies VTXO settlement contracts and oracle semantics. Per Arkana protocol, **human review is still required** before merge.

*🤖 Reviewed by Arkana — incremental review of commit `9d89ae6b7e..43f0503d48`*

ghost (commented):
# Arkana Code Review — arkade-os/compiler#28 (incremental, commit `b4ee33c`)

**Scope:** Single commit `43f0503..b4ee33c` — relaxes beacon height check from `>` to `>=`.

## Changes reviewed

| File | Change |
|---|---|
| `examples/stability/price_beacon.ark:50` | `newBlockHeight > currentHeight` → `newBlockHeight >= currentHeight` |
| `examples/stability/price_beacon.json` | **New file** — compiled artifact (was missing from prior commits) |
| `tests/beacon_test.rs:62` | Inline `PRICE_BEACON_CODE` fixture updated to match |
| `docs/stability-vault-prd.md:321` | PRD wording aligned: "monotonically non-decreasing" |

## Analysis

**Semantic change:** The oracle may now publish multiple price updates within the same Bitcoin block. Previously, strict `>` limited it to one update per block, contradicting the PRD's stated sub-block cadence design.

**Security assessment — LGTM:**

1. **No replay vector.** Each `update()` consumes the beacon UTXO. Even though the same block height can be reused, the UTXO model prevents replaying a stale price — the previous beacon is already spent.

2. **Staleness check unaffected.** Consumers enforce `tx.time - beaconHeight <= 144`. A beacon updated 5 times at height H is still "fresh as of H" — the staleness window is identical regardless of how many intra-block updates occurred.

3. **No cross-repo impact.** `PriceBeacon` is only consumed within the compiler repo (stability_vault.ark, stability_offer.ark). No downstream SDK or service references it yet.

4. **Settlement math unchanged.** `seekerRedeem`/`providerExit`/`take` all read the beacon's *price*, not its update frequency. The `>=` change doesn't alter any settlement formula.

5. **Consistency verified.** `.ark` source, inline test fixture, compiled JSON `source` field, and PRD all reflect the `>=` change.

## Observations (non-blocking)

1. **Missing compiled artifact.** `price_beacon.json` was absent in prior commits — good that it's added now, but this means the prior review rounds were working against a repo with an incomplete compilation artifact. Suggest running `generate_contracts.sh` as a CI check to prevent drift.

2. **Compiler type warnings.** The JSON includes 3 warnings about mixing `uint64le` / `scriptnum` types in comparisons. These are pre-existing (not introduced by this commit) but worth tracking — implicit type conversions in protocol-critical comparison opcodes are a class of bug I'd want explicit test coverage for.

3. **No runtime test for same-height updates.** The test suite validates compilation, not execution. A test that exercises two consecutive `update()` calls at the same block height would strengthen confidence. Low priority since this is a compiler repo, but worth adding when execution-level tests exist.

⚠️ **Protocol-critical reminder:** This PR modifies oracle monotonicity semantics — a core property that settlement contracts rely on. Per Arkana protocol, **human review is still required** before merge.

*🤖 Reviewed by Arkana — incremental review of commit `43f0503d48..b4ee33c`*

ghost (commented):
# Arkana Code Review — arkade-os/compiler#28 (incremental, commit `b4ee33c`)

**Scope:** Single commit `43f0503..b4ee33c` — relaxes beacon height check from `>` to `>=`.

## Changes reviewed

| File | Change |
|---|---|
| `examples/stability/price_beacon.ark:50` | `newBlockHeight > currentHeight` → `newBlockHeight >= currentHeight` |
| `examples/stability/price_beacon.json` | **New file** — compiled artifact (was missing from prior commits) |
| `tests/beacon_test.rs:62` | Inline `PRICE_BEACON_CODE` fixture updated to match |
| `docs/stability-vault-prd.md:321` | PRD wording aligned: "monotonically non-decreasing" |

## Analysis

**Semantic change:** The oracle may now publish multiple price updates within the same Bitcoin block. Previously, strict `>` limited it to one update per block, contradicting the PRD's stated sub-block cadence design.

**Security assessment — LGTM:**

1. **No replay vector.** Each `update()` consumes the beacon UTXO. Even though the same block height can be reused, the UTXO model prevents replaying a stale price — the previous beacon is already spent.

2. **Staleness check unaffected.** Consumers enforce `tx.time - beaconHeight <= 144`. A beacon updated 5 times at height H is still "fresh as of H" — the staleness window is identical regardless of how many intra-block updates occurred.

3. **No cross-repo impact.** `PriceBeacon` is only consumed within the compiler repo (stability_vault.ark, stability_offer.ark). No downstream SDK or service references it yet.

4. **Settlement math unchanged.** `seekerRedeem`/`providerExit`/`take` all read the beacon's *price*, not its update frequency. The `>=` change doesn't alter any settlement formula.

5. **Consistency verified.** `.ark` source, inline test fixture, compiled JSON `source` field, and PRD all reflect the `>=` change.

## Observations (non-blocking)

1. **Missing compiled artifact.** `price_beacon.json` was absent in prior commits — good that it's added now, but this means prior review rounds were against a repo with an incomplete compilation artifact. Suggest running `generate_contracts.sh` as a CI check to prevent drift.

2. **Compiler type warnings.** The JSON includes 3 warnings about mixing `uint64le` / `scriptnum` types in comparisons. These are pre-existing (not introduced by this commit) but worth tracking — implicit type conversions in protocol-critical comparison opcodes are a class of bug I'd want explicit test coverage for.

3. **No runtime test for same-height updates.** The test suite validates compilation, not execution. A test that exercises two consecutive `update()` calls at the same block height would strengthen confidence. Low priority since this is a compiler repo, but worth adding when execution-level tests exist.

⚠️ **Protocol-critical reminder:** This PR modifies oracle monotonicity semantics — a core property that settlement contracts rely on. Per Arkana protocol, **human review is still required** before merge.

*🤖 Reviewed by Arkana — incremental review of commit `43f0503d48..b4ee33c`*

ghost (approved):
# Arkana Code Review — arkade-os/compiler#28 (incremental, commit `f490853`)

**Scope:** Docs-only — `docs/stability-vault-prd.md` PriceBeacon section rewrite.

## What changed

The PriceBeacon design section was rewritten for clarity:

1. **Clock-as-block-height rationale** — now explicitly states *why* block height is used: `tx.time` maps to nLockTime block height, so both operands of the staleness check share the same unit. This was implicit before; making it explicit is good.

2. **Sub-block oracle cadence** — cleaner explanation that multiple intra-block updates share the same `clock` value (current block height) and that this is intentional, not a bug.

3. **Monotonicity enforcement** — `newBlockHeight >= currentHeight` (non-decreasing) spelled out more precisely, with explicit note that same-block updates are permitted while back-dating is rejected.

4. **Staleness window clarification** — notes that the 144-block window applies to the most recent published update, not individual intra-block updates.

## Verification

All PRD claims verified against contract code:
- `price_beacon.ark:update()` — `require(newBlockHeight >= currentHeight, ...)` ✓
- `stability_vault.ark` settlement functions — `tx.time - beaconHeight <= 144` ✓
- Unit consistency (block heights on both sides of arithmetic) ✓

## Verdict

Clean docs improvement. No code changes, no risk. **Approved.**

⚠️ **Protocol-critical reminder:** The overall PR modifies VTXO settlement contracts and oracle authentication. Per Arkana protocol, **human review is still required** before merge.

*🤖 Reviewed by Arkana — incremental review of commit `b4ee33c..f490853`*

ghost (approved):
# Arkana Code Review — arkade-os/compiler#28 (incremental, commit `f490853`)

**Scope:** Docs-only — `docs/stability-vault-prd.md` PriceBeacon section rewrite.

## What changed

The PriceBeacon design section was rewritten for clarity:

1. **Clock-as-block-height rationale** — now explicitly states *why* block height is used: `tx.time` maps to nLockTime block height, so both operands of the staleness check share the same unit. This was implicit before; making it explicit is good.

2. **Sub-block oracle cadence** — cleaner explanation that multiple intra-block updates share the same `clock` value (current block height) and that this is intentional, not a bug.

3. **Monotonicity enforcement** — `newBlockHeight >= currentHeight` (non-decreasing) spelled out more precisely, with explicit note that same-block updates are permitted while back-dating is rejected.

4. **Staleness window clarification** — notes that the 144-block window applies to the most recent published update, not individual intra-block updates.

## Verification

All PRD claims verified against contract code:
- `price_beacon.ark:update()` — `require(newBlockHeight >= currentHeight, ...)` ✓
- `stability_vault.ark` settlement functions — `tx.time - beaconHeight <= 144` ✓
- Unit consistency (block heights on both sides of arithmetic) ✓

## Verdict

Clean docs improvement. No code changes, no risk. **Approved.**

⚠️ **Protocol-critical reminder:** The overall PR modifies VTXO settlement contracts and oracle authentication. Per Arkana protocol, **human review is still required** before merge.

*🤖 Reviewed by Arkana — incremental review of commit `b4ee33c..f490853`*

PR #29, opened 2026-05-15
Add Lock-and-Mint contract suite example

tiero:
## Summary

Adds a complete contract suite to the examples directory, demonstrating a cross-chain token bridge implementation using Arkade. This includes four contracts (Endpoint, OApp, ReceiveMarker, SendMarker) with full source code, compiled JSON ABIs, and comprehensive integration tests.

## Key Changes

- **New Arkade contracts** (`examples/layerzero/`):
  - `endpoint.ark` — LayerZero Endpoint state machine with DVN-attested receive/send transitions
  - `oapp.ark` — USDT0 OApp with token mint/burn logic
  - `receive_marker.ark` — Receive-invocation marker (Endpoint→OApp)
  - `send_marker.ark` — Send-invocation marker (OApp→Endpoint)

- **Generated JSON ABIs** — Compiled output for all four contracts showing:
  - Constructor parameters (asset IDs, DVN pubkeys, exit delays)
  - Dual-path function variants (server-signed vs. exit-timelock fallback)
  - Full ASM emission with introspection opcodes
  - Compiler metadata and type warnings

- **Integration tests** (`tests/layerzero_test.rs`):
  - Validates all four contracts parse and compile successfully
  - Verifies DVN signature verification in Endpoint.receive()
  - Confirms asset-flow invariants (marker minting/burning, USDT0 delta checks)
  - Tests marker pinning to control singletons via `OP_INSPECTINASSETLOOKUP`
  - Validates Taproot introspection for state continuation

- **Documentation** (`examples/layerzero/README.md`):
  - High-level flow diagram showing Endpoint→OApp→Endpoint cycle
  - Asset role definitions and DVN attestation semantics
  - Boundary between Arkade contract enforcement and introspector runtime

- **Playground integration** — Updated `playground/main.js` to expose LayerZero contracts in the web IDE

## Implementation Details

The contracts enforce asset-level invariants (singleton control assets, marker minting/burning, USDT0 delta) via Arkade introspection opcodes (`OP_INSPECTINASSETLOOKUP`, `OP_FINDASSETGROUPBYASSETID`, `OP_INSPECTASSETGROUPSUM`). Packet-level invariants (LzReceivePacket parsing, nonce monotonicity, DVN attestation hash verification) are delegated to the introspector runtime layer, as documented in the accompanying README.

Both server-signed and exit-timelock fallback paths are implemented for all non-marker functions, following the dual-variant pattern established in other examples.

https://claude.ai/code/session_019K25jBMniDzNdzbdNuvEqQ

## Summary by CodeRabbit

* **New Features**
  * Added LayerZero / USDT0 contracts and documentation to the playground (Endpoint, OApp, ReceiveMarker, SendMarker).
  * Expanded the Arkade language with byte-string operations, packet inspection primitives, and `tx.id` plus Arkade input hash introspection.
* **Bug Fixes**
  * Prevented CLI input handling from prematurely consuming values, improving later artifact processing.
  * Improved detection of signature placeholders with case-insensitive matching.
* **Documentation**
  * Added an in-depth walkthrough for the LayerZero / USDT0 prototype and packet flow.
* **Tests**
  * Added comprehensive contract and opcode/ASM verification for the LayerZero examples and new primitives.


ghost (changes requested):
## Arkana Code Review — arkade-os/compiler#29

**Reviewer**: Arkana (aggressive review, protocol-level scrutiny)
**Verdict**: REQUEST CHANGES — 2 security findings, 2 design issues, several nits

---

### 🔴 CRITICAL — OApp.send() burn check allows silent USDT0 destruction

**`examples/layerzero/oapp.ark:87`**:
```
require(usdt0Group.sumInputs >= usdt0Group.sumOutputs + amount, "burn short");
```

The Go reference (`builders.go` ~line 1215-1237) enforces **exact equality**: `input_sum - output_sum == amountSD`. This contract uses `>=`, which means more USDT0 can be burned than the declared `amount`. A transaction could silently destroy USDT0 beyond what the OAppSendInvocation packet commits to.

The excess is irrecoverable — it's burned on-chain but not credited on the destination chain because the LzSend packet only carries `amount`, not the actual burn.

**Fix**: Change to `require(usdt0Group.delta == -amount, "burn mismatch")` (or the equivalent `sumInputs - sumOutputs == amount` using the compiler's delta semantics). Match the Go reference exactly.

---

### 🔴 HIGH — Marker exit variants are 0-of-0 multisig (anyone can grief after timelock)

**`examples/layerzero/receive_marker.json` (line ~61-71)** and **`send_marker.json` (line ~61-71)**:

Both marker contracts have zero pubkey parameters in constructor and zero in `consume()`. The compiler's `collect_all_pubkeys()` returns an empty set → the exit variant generates a **0-of-0 multisig**, meaning after the exit timelock, **anyone can spend the marker UTXO** with zero authorization.

Attack scenario: An attacker waits for the timelock to expire, then consumes (burns) a pending receive-invocation marker before `OApp.receive()` can process it. This griefs the cross-chain transfer — the DVN-attested inbound packet's marker is destroyed, and the credited USDT0 is never minted.

The Go reference (`BuildReceiveInvocationScript`, line 1333-1336) uses `OP_INSPECTINPUTARKADESCRIPTHASH` to pin the marker to the OApp's script hash, which is not available in the Arkade compiler grammar. But the exit variant should at minimum require *some* key to sign.

**Suggestion**: Add a `pubkey guardian` constructor parameter to both markers (could be the Endpoint or OApp operator key), so the exit variant becomes 1-of-1 instead of 0-of-0. Document that this is a stopgap until the compiler supports `OP_INSPECTINPUTARKADESCRIPTHASH`.

---

### 🟡 MEDIUM — OApp.send() ownerPk is witness-provided (no binding)

**`examples/layerzero/oapp.ark:81`**:
```
function send(int amount, signature ownerSig, pubkey ownerPk) {
    ...
    require(checkSig(ownerSig, ownerPk), "owner sig invalid");
```

`ownerPk` is a function parameter (witness data), not a constructor parameter. Anyone can provide their own keypair and satisfy this check. In the server-signed variant the server co-sign gates it, but in the **exit variant** (after timelock), an attacker provides their own `ownerPk` + `ownerPkSig` and can unilaterally burn USDT0 from the OApp.

The Go reference (`BuildOAppSendScript`) does NOT have this owner-sig check — it relies on the introspector runtime for sender authorization. If the intent is to add authorization beyond the server, `ownerPk` should be a constructor parameter (committed at contract creation) or bound to an on-chain value the introspector has verified.

---

### 🟡 MEDIUM — Endpoint.receive() missing sumInputs==0 on marker asset

**`examples/layerzero/endpoint.ark:68`**:
```
require(endpointIDGroup.sumOutputs == 1, "extra marker minted");
```

This only constrains the output side. If an unconsumed EndpointID marker from a previous receive happens to be included as an input (e.g., because OApp.receive() hasn't processed it yet), the net effect is the old marker is consumed and a new one minted — the output sum is still 1, so the check passes. This could allow "refreshing" a stale marker's packet data without a genuine DVN attestation for the new packet (the DVN sigs cover the `receiveHash` but the marker output carries whatever extension data the tx builder puts on it).

The Go reference also only checks `sumOutputs == 1` here, so this may be intentional. But adding `require(endpointIDGroup.sumInputs == 0, "unexpected marker input")` would be a defense-in-depth improvement. Worth a comment either way.

---

### 📝 Nits and observations

1. **Type warnings not addressed** — All four contracts produce `uint64le` vs `scriptnum` implicit-conversion warnings (visible in every `.json` file). The README acknowledges this implicitly but the contracts should use explicit `le64ToScriptNum()` calls, or the warnings should be documented as accepted with a rationale.

2. **JSON artifacts checked in without newline** — `endpoint.json`, `oapp.json`, `receive_marker.json`, `send_marker.json` all lack a trailing newline (`\ No newline at end of file`). Minor, but fails POSIX text file convention and can cause noisy diffs.

3. **playground/main.js references undeclared contract bindings** — The PR adds references to `contracts.endpoint`, `contracts.oapp`, `contracts.receive_marker`, `contracts.send_marker` but `contracts.js` is auto-generated by `generate_contracts.sh`. Should the PR also update that script, or does it already walk `examples/` recursively? If it walks recursively, this is fine — worth confirming.

4. **Test coverage is structural only** — Tests verify opcode presence in compiled ASM but don't test script execution with valid/invalid witnesses. No negative tests for `amount <= 0`, no test that the burn check actually rejects insufficient burns. This is acceptable for a first pass but should be tracked as follow-up.

5. **OApp.receive() hardcodes marker at input[0]** (`oapp.ark:38`) — The Go reference uses configurable `ReceiveInvocationInputIndex`. Fine for an example, but should be documented as a simplification.

---

### ✅ What looks good

- **DVN 2-of-2 attestation** in `endpoint.ark:receive()` is correctly implemented with `checkSigFromStack` over the canonical receive hash. Matches Go reference.
- **State continuation** pattern is solid — all stateful contracts (Endpoint, OApp) recursively check `tx.outputs[0].scriptPubKey == new Self(...)` with identical constructor params. Route preservation is enforced.
- **Marker lifecycle** is well-designed — mint-on-one-side, burn-on-other-side with group-sum checks.
- **Control-asset singleton** checks on both markers provide defense-in-depth binding to the consuming contract.
- **Documentation** (README.md) is excellent — the enforced-vs-delegated table is exactly the right way to communicate the boundary with the introspector layer.
- **Dual-variant pattern** (server + exit) follows established compiler conventions.
- **Test structure** is clean and follows existing patterns in the repo.

---

### Summary

The contract design is architecturally sound and the documentation is strong. The two critical findings (excess burn via `>=` and 0-of-0 marker exit variants) need to be fixed before merge. The ownerPk binding issue is a design decision worth discussing. This is protocol-adjacent code (asset minting/burning) — **requesting human review** per protocol-critical policy even after fixes are applied.

🤖 Reviewed by Arkana

ghost (changes requested):
## Arkana Code Review — arkade-os/compiler#29 (round 2, commits 318a116..fbee352)

**Reviewer**: Arkana (iterative review — new commits only)
**Verdict**: REQUEST CHANGES — 1 new critical finding, 1 prior critical partially unresolved, 2 new medium findings

---

### Prior findings status

| Finding | Severity | Status |
|---------|----------|--------|
| OApp.send() burn `>=` instead of `==` | 🔴 CRITICAL | ✅ **FIXED** — now `sumInputs == sumOutputs + bin2num(substr(tx.packet(20), 103, 8))` (exact equality) |
| Marker exit variants 0-of-0 multisig | 🔴 HIGH | ⚠️ **PARTIALLY FIXED** — see below |
| OApp.send() ownerPk witness-provided | 🟡 MEDIUM | ❌ Not addressed |
| Endpoint.receive() missing sumInputs==0 | 🟡 MEDIUM | ❌ Not addressed |

---

### ⚠️ Prior finding partially fixed: Marker exit variants still 0-of-0

The markers now use `OP_INSPECTINPUTARKADESCRIPTHASH` and `OP_PUSHCURRENTINPUTINDEX`, which are introspection ops. The compiler correctly detects this and generates an N-of-N exit path. **However**, both markers have zero `pubkey`-typed constructor/function parameters, so the N-of-N is literally 0-of-0:

`receive_marker.json` exit variant:
```json
"require": [{"type": "nOfNMultisig", "message": "0-of-0 signatures required (introspection fallback)"}],
"asm": ["<exit>", "OP_CHECKSEQUENCEVERIFY", "OP_DROP"]
```

After the CSV timelock expires, **anyone can spend the marker UTXO**. An attacker waits for the timelock, then burns a pending receive-invocation marker before `OApp.receive()` processes it → the cross-chain transfer is griefed.

The server-signed path is now properly secured (arkadeScriptHash binding is excellent). But the exit path remains trivially spendable.

**Fix (unchanged from round 1)**: Add a `pubkey guardian` constructor parameter to both markers so the exit variant becomes 1-of-1 + CSV. Or document that markers are expected to be consumed within the exit timelock window and the griefing is an accepted operational risk.

---

### 🔴 NEW — Missing `assertPacketAbsent` checks (packet injection attack surface)

**Files**: `endpoint.ark`, `oapp.ark` — all four functions

The Go reference (`builders.go`) explicitly rejects unexpected packet types in **every** script:
- `BuildEndpointReceiveScript` (line ~367-369): rejects `LzSend`, `OAppSendInvocation`
- `BuildEndpointSendScript` (line ~512-517): rejects `LzReceive`, `DvnAttestation`, `OAppSendInvocation`
- `BuildOAppReceiveScript` (line ~868-872): rejects ALL five packet types on current tx
- `BuildOAppSendScript` (line ~1147-1150): rejects `EndpointState`, `LzReceive`, `DvnAttestation`, `LzSend`

**None of the Arkade contracts perform these checks.**

The Go codebase has explicit comments explaining the risk: "Without this an attacker piggybacks a forged LzSend or OAppSendInvocation on a legitimate receive, hoping off-chain LayerZero infrastructure picks it up as a valid outbound message."

Attack scenario: Attacker constructs a legitimate `Endpoint.receive()` transaction but also attaches a forged `LzSend` packet. On-chain, the receive executes normally. Off-chain, LayerZero relayers see the piggybacked `LzSend` and process it as a valid outbound message → forged cross-chain transfer.

**Fix**: For each function, add `require(tx.packet(X) absent)` for every packet type that function should NOT see. This may require a new compiler primitive (e.g., `tx.packetAbsent(type)` that asserts OP_INSPECTPACKET returns 0), or invert the existing `tx.packet()` check. If the grammar doesn't support absence checks yet, this is a blocking gap.

---

### 🟡 NEW — Missing P2TR `0x5120` prefix verification in OApp.receive()

**`oapp.ark` receive()** — the recipient scriptPubKey pinning:
```
tx.outputs[1].scriptPubKey == substr(tx.inputs[0].packet(17), 147, 32)
```

This reads 32 bytes from the CreditMessage at offset 147 (after the `0x5120` P2TR prefix at offset 145). But it does **not verify the prefix is actually `0x5120`**. The Go reference (`builders.go:1026-1034`) explicitly checks the 2-byte P2TR tag.

If a malformed CreditMessage contains a non-P2TR prefix, funds would be sent to an unspendable or wrong-type output. DVN attestation mitigates this (the DVN wouldn't attest to a malformed packet), but the Go reference checks it as defense-in-depth.

**Fix**: Add `require(substr(tx.inputs[0].packet(17), 145, 2) == 0x5120, "not P2TR")` or equivalent.

---

### 🟡 STILL OPEN — OApp.send() `ownerPk` is witness-provided (round 1 finding)

**`oapp.ark` send()**:
```
function send(bytes32 sendMarkerScriptHash, signature ownerSig, pubkey ownerPk) {
    require(checkSig(ownerSig, ownerPk), "owner sig invalid");
```

`ownerPk` is a function parameter. Any keypair satisfies this check — it's security theater. The Go reference doesn't include this check at all. Either make `ownerPk` a constructor parameter (bound at contract creation) or remove the check entirely.

---

### ✅ What the new commits got right

**Compiler core is solid:**
- All 10 new opcodes match the introspector's canonical definitions (`OP_INSPECTPACKET`=0xf4, `OP_INSPECTINPUTPACKET`=0xf5, `OP_INSPECTINPUTARKADESCRIPTHASH`=0xc8, etc.)
- Grammar has no left-recursion or ambiguity issues — PEG ordered choice with distinct keyword prefixes on every alternative
- All emission sequences are stack-balanced: verified `OP_INSPECTPACKET` (pops type, pushes content+bool), `OP_INSPECTINPUTPACKET` (pops index then type, pushes content+bool), `OP_SUBSTR` (pops size, offset, data), etc.
- No `.unwrap()` or panics in new parser code — all error paths use `?` and `.ok_or()`
- Exit-path policy correctly enforced: `PacketInspect` and `InputPacketInspect` both return `true` from `expression_uses_introspection()`, and byte-string wrappers (Substr, Cat, etc.) propagate introspection status from children

**Packet field offsets verified correct:** All 35+ field offsets in the four contracts cross-referenced against Go constants — every single one matches.

**SHA256 bindings verified correct:**
- DVN attested hash: `sha256(LzReceive[1..141))` = 140 bytes covering receiver+srcEID+sender+nonce+GUID+messageHash ✓
- CreditMessage hash: `sha256(LzReceive[145..219))` vs `LzReceive[109..141)` ✓
- LzSend GUID: `sha256(OAppSendInvocation[0..175))` vs `LzSend[77..109)` ✓

**Marker arkadeScriptHash binding is excellent** — `ReceiveMarker.consume()` pins to OApp.receive() script hash, `SendMarker.consume()` pins to Endpoint.send() script hash, both with defense-in-depth singleton checks. This is a major improvement from round 1.

**10 new packet primitives tests** pin each opcode to its canonical name. Structural, but adequate for compiler wiring validation.

---

### 📝 Minor notes

1. **`Bin2Num` typed as `Uint64Le`** (`typechecker/mod.rs`) — `OP_BIN2NUM` produces a CScriptNum (variable-length LE signed integer), not a fixed LE64. Current typing is pragmatically functional (avoids conversion warnings when comparing to asset sums) but semantically incorrect. Worth a comment.

2. **Nonce monotonicity not enforced** — documented deferral in README, Go reference enforces it. Track as follow-up.

3. **No negative tests** — packet_primitives_test.rs and layerzero_test.rs only test happy paths. No tests for malformed inputs, wrong argument counts, or expected compilation failures.

---

### Summary

The two new commits are high quality: the compiler extensions are correct, well-tested, and the contract rewrite eliminates the runtime delegation. The original critical burn-check bug is fixed. The marker exit-path is improved but not fully resolved (still 0-of-0 after timelock). The new critical finding — missing packet-type rejection — is a real attack surface that the Go reference explicitly defends against. This needs to be addressed.

**Requesting human review** — this is protocol-critical code (asset minting/burning, cross-chain bridge logic).

🤖 Reviewed by Arkana

ghost (approved):
## Arkana Code Review — arkade-os/compiler#29 (round 3, commits fbee352..79aa582)
**Reviewer**: Arkana (iterative review — new commits only)

Two new commits since round 2. Both are clean.

---

### `00a562be` — fix(oapp): drop redundant owner signature in OApp.send()

**Security fix — correctly identified and resolved.** The removed `checkSig(ownerSig, ownerPk)` was a no-op: `ownerPk` was a witness parameter, so any caller could pass an arbitrary keypair with a valid self-signature. This authenticated nothing.

Verified the authority model post-fix:
- **Server variant**: all introspection checks (packet shape, route pinning, USDT0 burn via `sumInputs == sumOutputs + amount`, marker emission, state continuation) + server cosign. ✅
- **Exit variant**: CSV timelock only (`<exit> OP_CHECKSEQUENCEVERIFY OP_DROP`) — standard Ark unilateral exit pattern, same as `receive()`. ✅
- **USDT0 inputs**: each UTXO's per-owner script still requires its own signature. The OApp contract doesn't need to duplicate this. ✅
- **"0-of-0 signatures required"** label on exit variant is now accurate (was "1-of-1"). ✅

The updated test (`test_oapp_send_emits_send_marker`) is a strict improvement — it asserts zero `OP_CHECKSIG` before `<SERVER_KEY>` rather than merely checking one exists anywhere. Good regression guard.

Matches Go reference (`BuildOAppSendScript`) which has no owner sig check.

### `79aa582c` — fix(arkade-bindgen): clone cli.input before partial move

Mechanical Rust borrow-checker fix. `cli.input.expect()` partially moves the struct; newer rustc (1.94.x) rejects the subsequent `&cli` borrow. `.clone()` on `Option<PathBuf>` is negligible overhead. Correct fix. ✅

---

No protocol-critical concerns in this round. Both commits are bug fixes that improve correctness.

🤖 Reviewed by Arkana

ghost (approved):
## Arkana Code Review — arkade-os/compiler#29 (round 4, commits 79aa582..aa49711)
**Reviewer**: Arkana (iterative review — new commits only)

### New commits reviewed
- `d03468b` — CI rerun (no-op)
- `aa49711` — fix: add `operatorPk` so exit witness is non-empty

### Assessment: ✅ LGTM

This commit fixes a **critical security issue** that was caught by the new `compilation_roundtrip_test` from master: OApp, ReceiveMarker, and SendMarker had empty exit witnesses, meaning anyone could force-spend those VTXOs after the CSV timelock expired. Good catch by the CI sweep.

**Verified:**
1. **Exit ASM correctness** — All four contracts now have proper `<pk> <sig> OP_CHECKSIG <exit> OP_CHECKSEQUENCEVERIFY OP_DROP` (or CHECKSIGVERIFY chain + final CHECKSIG for Endpoint's 3-of-3). Stack discipline is correct: CHECKSIG leaves TRUE, CSV checks the exit value against nSequence without popping, DROP removes the exit value, script terminates with TRUE on top.

2. **State continuation consistency** — All `new Endpoint(...)`, `new OApp(...)`, `new ReceiveMarker(...)`, `new SendMarker(...)` calls in function bodies now include `operatorPk` in the correct parameter position (before `exit`). No parameter ordering mismatches.

3. **JSON ABI consistency** — Constructor params, VTXO template strings, witnessSchema, functionInputs, and nOfNMultisig messages all updated consistently across all four `.json` files.

4. **Cooperative path unchanged** — `operatorPk` does NOT appear in any function body logic (only in constructor params and `new ...()` continuations). The cooperative server-cosigned path remains DVN-attestation + introspection gated. This matches the Go reference.

5. **No cross-repo breakage** — Constructor param additions are backward-incompatible for anyone instantiating these contracts, but since this is an `examples/` directory and the contracts aren't consumed by downstream SDKs yet, this is fine.

**One note for the author** (non-blocking): The Endpoint now has a 3-of-3 exit (dvn0 + dvn1 + operator), while OApp/markers have 1-of-1 (operator only). This asymmetry is intentional and documented in the commit message — the Endpoint already had dvn keys, so they stay. Makes sense.

🤖 Reviewed by Arkana

tiero:
@arkanaai review

ghost:
## Review — Lock-and-Mint LayerZero/USDT0 contract suite

Solid PR overall. The compiler extensions are clean and the contracts faithfully mirror the Go reference scripts. A few things worth looking at before merge:

---

### 🔴 Potential bug: `scriptPubKey` vs x-only key in `OApp.receive()`

```ark
require(
  tx.outputs[1].scriptPubKey == substr(tx.inputs[0].packet(17), 147, 32),
  "recipient pkScript mismatch"
);
```

`OP_INSPECTOUTPUTSCRIPTPUBKEY` returns the full scriptPubKey bytes — for a P2TR output that's 34 bytes (`OP_1 0x20 <32-byte-key>`). The `substr` slice at `147..179` is only 32 bytes (the x-only key). These won't be equal unless the introspector has special P2TR handling that strips the 2-byte prefix.

The comment layout says:
```
0..2   P2TR script-pubkey tag (0x5120)   ← offset 145..147 in LzReceive
2..34  recipient x-only key              ← offset 147..179 in LzReceive
```

So the CreditMessage does carry the full tag + key starting at 145. Either:
1. Use `substr(packet, 145, 34)` for a full scriptPubKey match, **or**
2. Confirm that `OP_INSPECTOUTPUTSCRIPTPUBKEY` already strips the 2-byte P2TR witness-version prefix and returns only 32 bytes

Please clarify in a comment or the README, as this is the check that gates USDT0 delivery to the recipient.

---

### 🟡 Undeclared witness variables in `Endpoint.receive()`

`attestedHash`, `dvn0Sig`, and `dvn1Sig` are used in the function body but not declared as function inputs or `let` bindings. They must be implicitly added to the witness schema by the compiler (presumably via `checkSigFromStackVerify` argument inspection and the `sha256(...) == attestedHash` pattern).

The compiled JSON will reveal whether they appear in `witnessSchema`. If they do, all good — the pattern is actually elegant (prover provides `attestedHash`, and two independent `require`s pin it to both the sha256 of the LzReceive header and the DvnAttestation packet field). If they don't surface in the witness schema, the scripts would be unsatisfiable.

Worth adding a one-liner comment above the first use of `attestedHash` explaining it's a prover-supplied witness:
```ark
// attestedHash, dvn0Sig, dvn1Sig are witness elements supplied by the prover.
```

---

### 🟡 Nonce monotonicity deferral

The README does a good job explaining why it's deferred (`tx.inputs[currentInputIndex]` needs a literal index). That said, the deferred check means a malicious relayer could replay the same inbound nonce (as long as they have a valid DVN attestation for each replay). Is there a separate on-chain mechanism (e.g., state channel nonce in the Arkade round) that prevents this? Would be worth a sentence in the README on what the out-of-contract safety net is.

---

### ✅ What looks good

- **Compiler extensions** (`substr`, `cat`, `bin2num`, `num2bin`, `size`, `packet_inspect`, `input_packet_inspect`): clean ASM emission, correct `OP_SIZE OP_NIP` for `SizeOf`, correct `OP_INSPECTPACKET ... OP_1 OP_EQUALVERIFY` presence-assertion pattern.
- **Grammar PEG ordering**: `input_packet_inspect` before `packet_inspect` in alternatives — correct.
- **`hash_comparison` extension**: the `parse_additive_expr` path for complex preimages handles `sha256(substr(...)) == substr(...)` correctly. The fast-path for bare-identifier preimage + identifier RHS is preserved.
- **Validator relaxation**: allowing empty `witnessSchema` on exit variants is the right call for contracts with no constructor pubkeys.
- **`attestedHash` binding logic**: the three-way pin (`sha256(recv_header) == witness`, `packet18[1..32] == witness`, DVN sigs over `witness`) is a strong and correct construction.
- **Marker contracts**: `ReceiveMarker` and `SendMarker` correctly use `OP_INSPECTINPUTARKADESCRIPTHASH` to pin themselves to the consuming closure — this is the right primitive for invocation markers.
- **Tests**: opcode-presence assertions in `layerzero_test.rs` are a good pattern for pinning compiler output.
- **`arkade-bindgen` `.clone()` fix**: unrelated but correct, no issue.

---

One question on the `oappCtrlAssetId` field in `Endpoint`'s constructor — in the generated JSON it's a single `bytes32`, but `endpointCtrlAssetId`, `endpointIDAssetId`, etc. are split into `_txid` / `_gidx` pairs. Is that intentional (control asset ID encoded differently from ID assets), or a generation artifact?

The `scriptPubKey` size mismatch is the one I'd want confirmed before merging, since it directly affects whether the USDT0 credit actually lands at the intended recipient.

tiero:
Thanks @arkanaai — good catches, all three confirmed and fixed in `f0fb289`.

**1. Recipient `scriptPubKey` — bug confirmed.** Checked `docs/arkade-primitives-spec.md` Phase 7, which annotates `OP_INSPECTOUTPUTSCRIPTPUBKEY` as `outScript(bytes)` — a single 34-byte stack item (the full P2TR `0x5120 ‖ x-only-key`), not the `(program, version)` split that the public introspector README describes. So the LHS is 34 bytes and the old `substr(packet, 147, 32)` RHS was 32 bytes — never equal. Switched to `substr(tx.inputs[0].packet(17), 145, 34)`, which matches the full P2TR scriptPubKey including the `0x5120` tag. Added an inline comment pointing at the spec reference so the next person doesn't trip on the README ambiguity.

**2. Undeclared `attestedHash` / `dvn0Sig` / `dvn1Sig` — bug confirmed.** Inspected the compiled JSON: those identifiers appeared in `asm` as `<…>` placeholders but were absent from `witnessSchema`, so the cooperative script was indeed unsatisfiable. Added them as explicit `Endpoint.receive()` parameters; they now show up in `witnessSchema` as `attestedHash:bytes32`, `dvn0Sig:signature`, `dvn1Sig:signature`. The body's existing checks already pin `attestedHash` against both `sha256(substr(LzReceive, 1, 140))` and `substr(DvnAttestation, 1, 32)`, so the prover-supplied value stays tightly bound to canonical state. Added a "Witness convention" section to `examples/layerzero/README.md` documenting the pattern, with pointers to the analogous `htlc.claim(preimage)` and `fuji_safe.liquidate(currentPrice)` patterns.

**3. `oappCtrlAssetId` un-split — intentional artifact, now documented.** The compiler decomposes a `bytes32` constructor param into `_txid + _gidx` only when the function body passes it to `tx.{inputs,outputs}[i].assets.lookup(…)` or `tx.assetGroups.find(…)` (those need the `(txid32, gidx_u16)` pair the underlying `OP_INSPECT*ASSETLOOKUP` / `OP_FINDASSETGROUPBYASSETID` opcodes consume). `oappCtrlAssetId` is only passed through to `new ReceiveMarker(…)` inside Endpoint, so it stays as a single `bytes32`. Added a "Constructor decomposition" section to the README explaining the rule.

**Nonce monotonicity — expanded the README note.** You're right that the deferred on-chain `+1` check is the weakest link given just the on-chain story. Added the off-chain safety net: each DVN signs over the full LzReceive header, which already includes the inbound nonce as one of its bound fields. An on-chain replay would require a fresh DVN attestation over the *new* (replayed) header at the next nonce slot — which honest DVNs won't produce because the source-chain event is single-use and the DVN's signing rule binds `(srcEID, sender, dstEID, receiver, nonce)` to a specific emitted event. Combined with the route-prefix hash check that pins endpointID/oappID/route/DVN-keys, a nonce-field tamper would need a valid DVN attestation over the manipulated header. Still worth wiring the strict `+1` check on chain once `tx.inputs[currentInputIndex]` is exposed in the compiler grammar — flagged it as a small follow-up.

Suite: 282 passed, 0 failed (merged with master, includes the new bond tests).

---
_Generated by [Claude Code](https://claude.ai/code)_

tiero:
Thanks @coderabbitai — triage of the round-5 findings:

**Fixed in `25473b9` (this commit):**

- 🔴 `receive_marker.json` / `send_marker.json` `OP_EQUAL` operand order. Confirmed — the marker contracts were emitting `OP_PUSHCURRENTINPUTINDEX OP_EQUAL 0` instead of `OP_PUSHCURRENTINPUTINDEX 0 OP_EQUAL`. The `(Property, "==", Literal)` fast path in `generate_comparison_asm` had been emitting `<lhs> OP <rhs>` since master — dormant because every existing example reaches the comparison through `emit_binary_op_asm` or a dedicated rule (`time_comparison`, `group_property_comparison`, …) which all emit the correct push-push-op order. The LayerZero markers are the first real consumer of this branch. Flipped to the correct order; suite is 282/282.

**Already addressed in `f0fb289`:**

- 🔴 `endpoint.ark:155` — undefined `attestedHash`/`dvn0Sig`/`dvn1Sig`. Declared as function parameters (`receive(bytes32 receiveMarkerScriptHash, bytes32 attestedHash, signature dvn0Sig, signature dvn1Sig)`) so they surface in `witnessSchema`. The two existing on-chain checks (`sha256(substr(LzReceive, 1, 140)) == attestedHash` and `substr(DvnAttestation, 1, 32) == attestedHash`) still pin the witness to canonical packet-derived values — same pattern as `htlc.claim(preimage)`. The `let attestedHash = substr(packet18, 1, 32)` alternative would also work but loses the explicit "this is a prover-supplied witness, here are the on-chain pins" shape.

**Skipped with rationale:**

- 🟠 `src/parser/mod.rs:935` — `sha256_func` matching only specific child rules vs `Rule::additive_expr`. Verified empirically: every LayerZero contract's compiled JSON inlines `OP_SHA256` correctly (`endpoint.json` has zero `<sha256(...)>` placeholders). The concern is theoretically valid for some grammar paths but doesn't surface in any contract in this PR.

- 🟠 `src/compiler/mod.rs:1490-1580` — other `Property` comparison branches still emit `<this.*>` placeholders. Only `(Property, "==", Literal)` is hit by my code; the sibling branches `(Property, ">=", Literal)`, `(Property, ==, Variable)`, etc. are unused by the LayerZero contracts and remain dormant in the same way the master examples don't hit them. Flagged as a follow-up cleanup rather than fixing in this PR — the broken-order convention is wider than the `this.*` mapping and warrants its own scoped change.

- 🟠 `src/parser/mod.rs:2091-2110` — `parse_constructor_args` falling back to `Variable` for unknown rules. Not exercised by the LayerZero contracts (all `new …(…)` args are bare identifiers or constructor params). Follow-up.

**Nitpicks intentionally not taken on:**

- `examples/layerzero/README.md` fence missing language tag — cosmetic, deferring.
- `src/validator/mod.rs` warning split unit tests, `tests/layerzero_test.rs` exit-variant assertions — reasonable additions but tightening test coverage is out of scope for this round.

Suite: 282 passed, 0 failed.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/parser/mod.rs
```diff
@@ -799,14 +866,138 @@ fn parse_hash_comparison(pair: Pair<Rule>) -> Result<Requirement, String> {
...
+                Rule::input_introspection => parse_input_introspection_to_expression(arg_pair)?,
+                Rule::output_introspection => parse_output_introspection_to_expression(arg_pair)?,
+                Rule::identifier => Expression::Variable(arg_pair.as_str().to_string()),
+                Rule::number_literal => Expression::Literal(arg_pair.as_str().to_string()),
+                _ => Expression::Property(arg_pair.as_str().to_string()),
+            };
+            Ok(Expression::Sha256 {
+                data: Box::new(data),
+            })
+        }
```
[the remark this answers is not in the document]
tiero: Done in 56bf9c1 — `sha256()`'s argument now routes through `parse_additive_expr`, so `sha256(substr(...))` emits inline OP_SUBSTR+OP_SHA256 instead of a `<sha256(...)>` placeholder. Added a regression test in packet_primitives_test.rs.

examples/layerzero/receive_marker.json
```diff
@@ -0,0 +1,103 @@
...
+          "type": "assetCheck"
+        },
+        {
+          "type": "serverSignature"
+        }
+      ],
+      "asm": [
+        "OP_PUSHCURRENTINPUTINDEX",
+        "OP_EQUAL",
+        "0",
```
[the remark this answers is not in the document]
tiero: Fixed — `Property == Literal` now emits left, right, OP_EQUAL (25473b9), and 56bf9c1 additionally makes the `require(expr)` => `expr == true` dummy emit a bare introspection push. Marker JSONs regenerated and a regression test added.

examples/layerzero/endpoint.ark
```diff
@@ -0,0 +1,285 @@
...
+    require(substr(tx.packet(18), 132, 32) == dvn1Pk, "dvn att1 pk mismatch");
+    require(
+      sha256(substr(tx.packet(17), 1, 140)) == attestedHash,
+      "attested hash does not match lz receive header"
+    );
+    require(substr(tx.packet(18), 1, 32) == attestedHash, "dvn attested hash mismatch");
+
+    // DVN 2-of-2 signatures over the attestedHash.
+    require(checkSigFromStackVerify(dvn0Sig, dvn0Pk, attestedHash), "dvn0 sig invalid");
+    require(checkSigFromStackVerify(dvn1Sig, dvn1Pk, attestedHash), "dvn1 sig invalid");
```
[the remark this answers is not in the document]
tiero: Already addressed in f0fb289 — `receive()` declares `dvn0Sig`/`dvn1Sig` as `signature` params and extracts `attestedHash` from the DVN packet, so the witnesses bind at construction time.

tiero:
Thanks @coderabbitai. I've opened #39 for item **#2** (the reversed operand order in the remaining `generate_comparison_asm` sibling branches) — no need to open a duplicate for that one. Items **#1** (sha256_func grammar match) and **#3** (parse_constructor_args unknown-rule fallback) are already mitigated in this PR / dormant; I'll fold them into #39 or file separately if they resurface, so please don't auto-open issues for those either.

tiero:
Havent reviewed the opcodes (as not the goal of this PR to generate correct bytecode) rather to look into semantics / grammar @msinkec

examples/layerzero/endpoint.ark
```diff
@@ -0,0 +1,297 @@
...
+  //   - Output[1] uses the canonical ReceiveMarker pkScript.
+  // -------------------------------------------------------------------------
+  function receive(
+    bytes32 receiveMarkerScriptHash,
+    bytes32 attestedHash,
+    signature dvn0Sig,
+    signature dvn1Sig
+  ) {
+    // Pull the three current-tx packets. tx.packet() asserts presence.
+    require(size(tx.packet(16)) == 183, "endpoint state packet size");
```
msinkec, line 132: Would be nicer with global compile time constants, e.g. const int ENDPOINT_STATE_PACKED_ID = 16 and so on once supported.

examples/layerzero/endpoint.ark
```diff
@@ -0,0 +1,297 @@
...
+    require(substr(tx.packet(18), 35, 32) == dvn0Pk, "dvn att0 pk mismatch");
+    require(substr(tx.packet(18), 132, 32) == dvn1Pk, "dvn att1 pk mismatch");
+    require(
+      sha256(substr(tx.packet(17), 1, 140)) == attestedHash,
+      "attested hash does not match lz receive header"
+    );
+    require(substr(tx.packet(18), 1, 32) == attestedHash, "dvn attested hash mismatch");
+
+    // DVN 2-of-2 signatures over the attestedHash.
+    require(checkSigFromStackVerify(dvn0Sig, dvn0Pk, attestedHash), "dvn0 sig invalid");
```
msinkec: We have no OP_CHECKSIGFROMSTACKVERIFY, so I assume this would compile down to OP_CHECKSIGFROMSTACK OP_VERFIY ? I don't think we need that as a builtin. Besides it would fail early, not via the require(). Shold be simply:

`require(checkSigFromStack(dvn0Sig, dvn0Pk, attestedHash), "dvn0 sig invalid");`

tiero: agreed

examples/layerzero/endpoint.ark
```diff
@@ -0,0 +1,297 @@
...
+    // marker to the body OApp.receive() will read.
+    require(
+      sha256(substr(tx.packet(17), 145, 74)) == substr(tx.packet(17), 109, 32),
+      "credit message hash mismatch"
+    );
+
+    // Endpoint state continuation: output[0] uses the same pkScript and
+    // carries the Endpoint control asset.
+    require(
+      tx.outputs[0].scriptPubKey == new Endpoint(
```
msinkec: Constructor params usually stay the same in recursive covenants. We could introduce something like new self() to simplify the code.

examples/layerzero/endpoint.ark
```diff
@@ -0,0 +1,297 @@
...
+        endpointCtrlAssetId, endpointIDAssetId,
+        oappCtrlAssetId,     oappIDAssetId,
+        endpointID, oappID, remoteEID, remoteOApp,
+        dvn0Pk, dvn1Pk, exit
+      ),
+      "endpoint state must continue"
+    );
+    require(
+      tx.outputs[0].assets.lookup(endpointCtrlAssetId) == 1,
```
msinkec: Lookup opcode requires the gidx of the asset to also be passed.

msinkec: I think lookup() is meant to decompose the asset id internally? This is better to be a seperate input as it may be dynamic and not known at compile time.

examples/layerzero/endpoint.ark
```diff
@@ -0,0 +1,297 @@
...
+    );
+    require(
+      tx.outputs[1].scriptPubKey == new ReceiveMarker(
+        receiveMarkerScriptHash, oappCtrlAssetId, exit
+      ),
+      "marker pkScript not canonical"
+    );
+
+    // Exactly one EndpointID asset minted (no extras).
+    let endpointIDGroup = tx.assetGroups.find(endpointIDAssetId);
```
msinkec: Same here with group idx.

examples/layerzero/endpoint.ark
```diff
@@ -0,0 +1,297 @@
...
+    require(endpointIDGroup.sumOutputs == 1, "extra marker minted");
+  }
+
+  // -------------------------------------------------------------------------
+  // ENDPOINT SEND
+  // Consume the OApp-emitted send-invocation marker and emit an outbound
+  // LzSendPacket. Endpoint state continues; the OAppID marker is fully burned.
+  // -------------------------------------------------------------------------
+  function send() {
+    require(size(tx.packet(16)) == 183, "endpoint state packet size");
```
msinkec, line 223: For the sake of readability we could assign packets to variables.
`let endpointStatePacket = tx.packet(16);`

examples/layerzero/endpoint.ark
msinkec: Actually, now that I think about it we can just do:
`tx.outputs[0].scriptPubKey == tx.input.current.scriptPubKey`

examples/layerzero/endpoint.ark
tiero: Fixed in `961131e` — both DVN checks now use `require(checkSigFromStack(dvn*Sig, dvn*Pk, attestedHash), "...")`, emitting plain `OP_CHECKSIGFROMSTACK` (0xcc, the only variant the introspector has). The `checkSigFromStackVerify` builtin is gone from the contract; the failure now flows through the require() path instead of aborting early. Test updated to count `OP_CHECKSIGFROMSTACK` occurrences.

---
_Generated by [Claude Code](https://claude.ai/code)_

examples/layerzero/endpoint.ark
tiero: Adopted in `961131e` — all four state continuations (Endpoint.receive/send, OApp.receive/send) now use `tx.outputs[0].scriptPubKey == tx.input.current.scriptPubKey`, which compiles to `OP_INSPECTOUTPUTSCRIPTPUBKEY OP_PUSHCURRENTINPUTINDEX OP_INSPECTINPUTSCRIPTPUBKEY OP_EQUAL` — byte-for-byte the state-continuation check in the Go reference (`BuildEndpointReceiveScript`). It's also a stronger guarantee than re-deriving the constructor (script equality vs parameter-set equality) and removes the eleven-arg `new Endpoint(...)` boilerplate.

The `new ReceiveMarker(...)` / `new SendMarker(...)` pins stay in constructor form since those target a *different* contract's pkScript. A `new self()` sugar would still be nice for the cases where a param does change across iterations (e.g. `PriceBeacon.migrate` rotating `oraclePk`) — worth folding into the constants/ergonomics follow-up.

---
_Generated by [Claude Code](https://claude.ai/code)_

examples/layerzero/endpoint.ark
tiero: Right — `lookup()` decomposes internally, and the gidx *is* a separate dynamic input. The compiler splits any `bytes32` constructor param used in `assets.lookup(...)` / `assetGroups.find(...)` into two ABI-level constructor inputs: `endpointIDAssetId_txid` (bytes32) + `endpointIDAssetId_gidx` (int). The emitted asm is `<endpointIDAssetId_txid> <endpointIDAssetId_gidx> OP_INSPECTOUTASSETLOOKUP`, so both halves are placeholders resolved at instantiation time — nothing is baked at compile time. Params that are only passed through to a child constructor stay as a single bytes32.

This is documented in the README's "Constructor decomposition" section (added after Arkana asked the same question). If we'd rather make the split visible at the source level (e.g. an explicit `asset` type that desugars to the pair), that's a nice candidate for the same ergonomics follow-up as `const`.

---
_Generated by [Claude Code](https://claude.ai/code)_

examples/layerzero/endpoint.ark
tiero: Same answer as the lookup thread above: `tx.assetGroups.find(endpointIDAssetId)` decomposes into `<endpointIDAssetId_txid> <endpointIDAssetId_gidx> OP_FINDASSETGROUPBYASSETID`, with both halves as separate instantiation-time constructor inputs in the ABI (`endpointIDAssetId_txid: bytes32`, `endpointIDAssetId_gidx: int`). Nothing is fixed at compile time — see the README "Constructor decomposition" section.

---
_Generated by [Claude Code](https://claude.ai/code)_

examples/layerzero/endpoint.ark
tiero: Agreed on the readability win, but deferring: today's `let` handling emits the bound expression at the binding site without stack tracking, so a later `substr(endpointStatePacket, …)` would emit an unresolvable `<endpointStatePacket>` placeholder rather than re-inspecting the packet. Making `let pkt = tx.packet(16)` work needs either compiler-side inline substitution of let-bound byte expressions or real stack-slot tracking — a language follow-up alongside `const`. Note the repeated `tx.packet(16)` calls mirror the Go reference, which also re-pushes the packet via `OP_INSPECTPACKET` for every field check.

In the meantime `961131e` adds a header note mapping the packet-type ids (16=EndpointState … 20=OAppSendInvocation) so the literals are at least documented at the top of the file.

---
_Generated by [Claude Code](https://claude.ai/code)_

examples/layerzero/endpoint.ark
tiero: Agreed — `const int ENDPOINT_STATE_PACKET_ID = 16;` would be the right shape once the language supports it. As an interim step `961131e` adds a header note in this file mapping every packet-type literal (16=EndpointState 0x10, 17=LzReceive 0x11, 18=DvnAttestation 0x12, 19=LzSend 0x13, 20=OAppSendInvocation 0x14) with an explicit pointer to hoist them into `const` declarations when available. Tracking the `const` language feature in the same ergonomics follow-up as `let`-bound packets and `new self()`.

---
_Generated by [Claude Code](https://claude.ai/code)_

tiero:
@coderabbitai — triage of the round-6 findings (commits through `d90f1f3`):

**Fixed in `cc99cf7`:**

- ✅ `src/validator/mod.rs` — signature-leak self-check case sensitivity. Confirmed: `t.contains("Sig>")` missed lowercase placeholders like `<sig>` (this module's own fixture naming) and `<ownersig>`, silently defeating the witness-only invariant. Switched to `t.to_ascii_lowercase().contains("sig>")` and added a regression test covering `<serverSig>`, `<ownersig>`, and `<sig>`. Legit leaf-asm tokens are unaffected — key placeholders end in `_KEY>` / `:fn>`. Suite: 341 passed, 0 failed.

**Rejected — stale coding guideline (4 findings):**

- ⛔ `endpoint.ark` / `oapp.ark` / `receive_marker.ark` / `send_marker.ark` — "add `options { server = server; exit = exit; }` and `int exit`". This guideline predates the tapscript refactor merged from master (PR #50, now in this branch via `d90f1f3`). **The `options` block was removed from the language entirely** — `src/parser/grammar.pest` has zero occurrences of `options`; the proposed fix would be a parse error. In the new model, cooperative signing is a synthesized default collaborative leaf (`checkMultisig([server, tweak(emulator, fn)], …, 2)`) per covenant function, and unilateral exit is an explicit standalone CSV `tapscript` leaf — the four LayerZero contracts follow the migrated pattern intentionally. Please refresh the `examples/**/*.ark` guideline to the post-#50 conventions so this doesn't re-fire on every example PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #33, opened 2026-05-21
Add CoveredCall and CashSecuredPut option contracts with tests

tiero:
## Summary
Adds two new physically-settled European option contracts inspired by Rysk Finance v12: `CoveredCall` and `CashSecuredPut`. Both contracts support dual-path execution (cooperative server-assisted and exit timelock fallback) and allow either party to transfer their position before expiry.

### New Contracts
- **`examples/covered_call.ark`**: Seller locks BTC collateral; buyer exercises by paying strike in stablecoin to receive the BTC within the exercise window. After expiry, seller reclaims.
- **`examples/cash_secured_put.ark`**: Seller locks stablecoin collateral; buyer exercises by delivering BTC to receive the stablecoin within the exercise window. After expiry, seller reclaims.

### Generated Artifacts
- **`examples/covered_call.json`** and **`examples/cash_secured_put.json`**: Compiled ABI + script assembly for both contracts, each with 8 taproot leaves (4 functions × 2 variants: server-assisted and exit timelock).

### Test Coverage
- **`tests/covered_call_test.rs`**: 8 integration tests validating:
  - Compilation to 8 tapleaves
  - Exercise uses asset introspection (no oracle)
  - Exercise is buyer-gated
  - Reclaim is seller-only
  - Transfer functions preserve collateral
  - Dual-path variants (server and exit)

- **`tests/cash_secured_put_test.rs`**: 8 integration tests with parallel coverage for the put variant.

## Implementation Details

- Both contracts follow the established pattern: `options { server = server; exit = exit; }` with constructor parameters for `int exit`.
- Exercise windows are defined as `[expiryHeight, expiryHeight + graceBlocks)`.
- Asset collateral is verified via `OP_INSPECTOUTASSETLOOKUP` (no oracle dependency).
- Transfer functions (`transferSeller`, `transferBuyer`) are pure key swaps that preserve collateral and contract terms.
- Dual-path execution: server variant uses cooperative co-signature; non-server variant uses 3-of-3 multisig with exit sequence lock.
- Compiler generates appropriate warnings for type conversions (uint64le vs scriptnum) and output-invariant placeholders.

https://claude.ai/code/session_01MXpCrecXFzQ3bmHqUuW5yx


ghost (changes requested):
## 🔍 Code Review — arkade-os/compiler#33

**Reviewer:** Arkana (automated)
**Verdict:** Request changes — protocol-critical contract code, requires human sign-off regardless

---

### ⚠️ PROTOCOL-CRITICAL: These contracts lock real BTC and stablecoins. Human review required before merge.

---

### 🔴 HIGH — Exercise/Transfer race in cooperative path

**`covered_call.ark:89-100`, `cash_secured_put.ark:103-117`**

`transferSeller` has no timelock constraint. During the exercise window `[expiryHeight, expiryHeight + graceBlocks)`, both `exercise` (buyer) and `transferSeller` (seller) are valid spending paths. A malicious seller can front-run a buyer's exercise by cooperating with the server to transfer the contract to a new seller key they control — effectively revoking the buyer's ability to exercise against the original UTXO.

In the **server-assisted path**, the server is the mediator and could refuse conflicting transfers. But this is a trust assumption that should be documented. In the **exit path**, both require CSV, so the race is between multisig cooperation, which is less exploitable.

**Recommendation:** Either (a) add a time guard preventing transfers during the exercise window (`require(tx.time < expiryHeight || tx.time >= expiryHeight + graceBlocks)`), or (b) document explicitly that the server MUST refuse transfers during the exercise window. Option (a) is strictly safer.

---

### 🟡 MEDIUM — No validation of degenerate constructor parameters

**Both contracts, constructor parameters**

No constraints on constructor inputs:
- `graceBlocks = 0` → exercise window is `[expiryHeight, expiryHeight)` — empty range. Option is unexercisable. Seller collects premium and reclaims immediately at expiry. This is a griefing vector if the buyer doesn't validate before funding.
- `btcSats = 0` / `strikeAmount = 0` / `stableAmount = 0` → degenerate options with no economic value but valid contract state.
- `expiryHeight` in the past → immediately reclaimable, exercise window already closed.

This is typically SDK-level validation, but given these are reference examples, consider adding a comment documenting the assumption that the SDK/buyer MUST validate these before funding.

---

### 🟡 MEDIUM — Compiler output-invariant warnings indicate unresolvable placeholders

**`covered_call.json` warnings, `cash_secured_put.json` warnings**

The compiled JSON contains 10 warnings each, including:

```
warning[output-invariant]: fn 'exercise' (serverVariant=true): placeholder <tx.time > is not in witnessSchema or constructorInputs
warning[output-invariant]: fn 'exercise' (serverVariant=true): placeholder <windowClose> is not in witnessSchema or constructorInputs
warning[output-invariant]: fn 'reclaim' (serverVariant=true): placeholder <reclaimHeight> is not in witnessSchema or constructorInputs
```

Two concerns:
1. **Trailing space in `<tx.time >`** — looks like a tokenizer/parser bug in the compiler. The ASM contains `<tx.time >` instead of `<tx.time>`.
2. **`<windowClose>` and `<reclaimHeight>` are computed locals** (from `expiryHeight + graceBlocks`) that appear as unresolvable placeholders in the ASM. The SDK must know how to resolve these at transaction construction time. Is this documented? Are there existing SDK tests that exercise this pattern?

These warnings suggest the generated artifacts may not be directly consumable by SDKs without additional binding logic. Confirm the ts-sdk / go-sdk can handle these placeholders before merging.

---

### 🟡 MEDIUM — `reclaim` has zero output constraints

**`covered_call.ark:82-85`, `cash_secured_put.ark:92-96`**

The `reclaim` function only enforces timelock + seller signature. There are no output constraints — the seller can send the collateral anywhere. This is correct by design (seller reclaims their own collateral), but it's worth noting that if the Ark server cooperates with a reclaim BEFORE the exercise window closes (e.g., server bug), there's no script-level protection for the buyer.

The buyer's protection relies entirely on the CLTV enforcement. This is sound as long as `expiryHeight + graceBlocks` is correctly computed at the script level. Given the `<reclaimHeight>` placeholder warning above, confirm this CLTV value is correctly bound.

---

### 🔵 LOW — Test coverage gaps

**`tests/covered_call_test.rs`, `tests/cash_secured_put_test.rs`**

Tests are structural (verify opcode presence, function counts, input names). Missing:
- **Boundary tests**: What happens at exactly `expiryHeight` and `expiryHeight + graceBlocks`?
- **Degenerate inputs**: `graceBlocks = 0`, zero amounts
- **Test naming inconsistency**: `test_transfer_seller_preserves_collateral` and `test_transfer_buyer_preserves_collateral` in `covered_call_test.rs` assert "no asset lookup" (`!asm.contains(OP_INSPECTOUTASSETLOOKUP)`), but the contract DOES check `tx.outputs[0].value >= btcSats` via `OP_INSPECTOUTPUTVALUE`. The test names say "preserves collateral" but the assertions only verify "no asset lookup" — they don't verify the `OP_INSPECTOUTPUTVALUE` check IS present. This is a false-confidence test.
- **No cross-function interference tests**: Can a buyer exercise AND a seller reclaim reference the same UTXO?

---

### 🔵 LOW — Committed generated artifacts

**`examples/covered_call.json`, `examples/cash_secured_put.json`**

These 1138 lines of JSON are compiler output. Committing them means:
- They can drift from the `.ark` source if someone edits the source but forgets to regenerate
- They inflate the diff (1138 of 1763 added lines are generated)

Consider: generate on CI, or add a `make verify` target that confirms `.json` matches `.ark`.

---

### ✅ What looks good

- **Contract logic is fundamentally sound**: Exercise window boundaries are clean (strict `<` for close, `>=` for open). No overlap between exercise and reclaim.
- **Dual-path execution**: Server-assisted (cooperative + introspection) and exit (N-of-N multisig + CSV) paths follow established compiler patterns.
- **Asset introspection**: `OP_INSPECTOUTASSETLOOKUP` for stablecoin verification, `OP_INSPECTOUTPUTVALUE` for BTC verification — correct use of Elements opcodes.
- **Transfer functions preserve all terms**: Constructor params are correctly threaded through `new CoveredCall(...)` / `new CashSecuredPut(...)` on transfer outputs.
- **No oracle dependency**: Physical settlement via buyer's exercise decision is a clean design.
- **Consistent with existing patterns**: Follows `stability_offer.ark` and `non_interactive_swap.ark` patterns.

---

### Summary

| Severity | Count | Action |
|----------|-------|--------|
| 🔴 HIGH | 1 | Transfer/exercise race needs mitigation |
| 🟡 MEDIUM | 3 | Parameter validation, placeholder warnings, reclaim constraints |
| 🔵 LOW | 2 | Test gaps, committed artifacts |

**Bottom line:** The contract design is solid and the code is well-structured. The HIGH finding (transfer/exercise race) needs resolution — either a time guard on transfers or explicit server policy documentation. The MEDIUM findings are worth addressing before this graduates from example to production.

🏷️ **Protocol-critical: requires human approval before merge.**

🤖 Reviewed by Arkana

ghost (changes requested):
## Arkana Code Review — `CoveredCall` & `CashSecuredPut`

**⚠️ Protocol-critical: option contracts handling BTC and stablecoin settlement. Requires human sign-off.**

Good contract design overall — exercise is buyer-gated, reclaim is seller-gated, transfers preserve all terms, dual-path follows Arkade patterns correctly, overflow protection on `expiryHeight + graceBlocks` via `OP_ADD64 + OP_VERIFY`, and the docs (options.md) are thorough and honest about limitations. The concerns below are real and need to be addressed before merge.

---

### 🔴 P0 — Transfer functions lack exercise-window guards

**Files:** `examples/covered_call.ark:97-110`, `examples/cash_secured_put.ark:91-109`

`transferSeller` and `transferBuyer` have no timelock restriction. In the **cooperative path**, a seller can ask the Operator to co-sign a transfer during the exercise window, consuming the vault UTXO and **invalidating the buyer's pre-signed exercise template**.

The exit path is safe (requires buyer's signature). But the cooperative path relies entirely on Operator policy — nothing in the script prevents it.

**Impact:** Seller griefing vector. A seller who sees the option going deep ITM can transfer the vault to a fresh key, destroying the buyer's pre-signed exercise tx. The buyer has to re-negotiate exercise with the new "seller" (who is the same person with a new key).

**Recommendation:** Add `require(tx.time < expiryHeight, "transfers locked at expiry")` to both transfer functions. Yes, this has the compiler limitation noted in P1 — but even as documentation-in-code that the Operator must enforce, it's better than nothing. Alternatively, add a comment block explicitly stating the Operator MUST reject co-signing transfers at/after `expiryHeight`.

---

### 🟡 P1 — `tx.time < windowClose` compiles to non-functional ASM (compiler limitation)

**Files:** `examples/covered_call.ark:58`, `examples/cash_secured_put.ark:37`

The compiler grammar only handles `tx.time >= identifier` as a first-class pattern (→ `OP_CHECKLOCKTIMEVERIFY`). The `tx.time < windowClose` comparison falls through to generic `property_comparison`, emitting:

```
"<tx.time >",    ← note trailing space — parser bug
"<windowClose>",  ← local variable, unresolvable
"OP_LESSTHAN",
```

These are non-functional placeholders. In the cooperative path the Operator validates this off-chain, so the contract works. But the compiled script contains dead code that **cannot execute** and the upper-bound check exists only at the Operator trust level.

This is the first contract in the repo to use `tx.time <` — every other contract (htlc, swap, payment_auth, fuji_safe) only uses `tx.time >=`.

**Recommendation:** Either (a) file a compiler issue for `tx.time <` support, or (b) restructure to use only `tx.time >=` patterns (e.g., compute and pass `windowClose` as a constructor parameter, use `tx.time >= windowClose` as a "reclaim is valid" signal instead of "exercise is invalid"). At minimum, add a code comment explaining that this check is Operator-enforced only.

---

### 🟡 P2 — No minimum `graceBlocks` validation

**Files:** `examples/covered_call.ark:41`, `examples/cash_secured_put.ark:27`

If instantiated with `graceBlocks = 0`, the exercise window `[expiryHeight, expiryHeight + 0)` is empty — the buyer can **never** exercise. With `graceBlocks = 1`, the window is one block — practically impossible to hit.

Neither the contract nor the compiler warns about this. A misconfigured deployment silently creates an unexercisable option (seller keeps collateral + premium for free).

**Recommendation:** Add `require(graceBlocks >= 144, "grace period too short")` in the constructor or document a minimum in options.md. 144 blocks ≈ 1 day is the standard Arkade safety margin.

---

### 🟡 P3 — `<reclaimHeight>` placeholder in reclaim ASM is unresolvable

**Files:** `examples/covered_call.json` (reclaim variants), `examples/cash_secured_put.json` (reclaim variants)

The reclaim functions compute `reclaimHeight = expiryHeight + graceBlocks` as a local variable. The compiled ASM correctly computes this via `OP_ADD64` but then uses `<reclaimHeight>` as the `OP_CHECKLOCKTIMEVERIFY` argument:

```
"<expiryHeight>", OP_SCRIPTNUMTOLE64, "<graceBlocks>", OP_SCRIPTNUMTOLE64,
OP_ADD64, OP_VERIFY,
"<reclaimHeight>",      ← unresolvable placeholder
OP_CHECKLOCKTIMEVERIFY, OP_DROP,
```

The `OP_ADD64` result (the actual computed value) is still on the stack but isn't used by CLTV — instead, a separate unresolvable `<reclaimHeight>` placeholder is pushed. The SDK consuming this artifact would need to know to substitute `expiryHeight + graceBlocks` for `<reclaimHeight>`, but this isn't documented anywhere.

**Recommendation:** This is a compiler codegen issue — the CLTV should consume the ADD64 result from the stack rather than pushing a separate placeholder. File a compiler bug. For now, document in the JSON warnings or in options.md that `<reclaimHeight> = expiryHeight + graceBlocks`.

---

### 🟢 P4 — Nits

1. **JSON files missing trailing newline** — `covered_call.json:562`, `cash_secured_put.json:576`. Minor but causes noisy diffs.

2. **Test coverage is structural only** — Tests verify opcode presence and tapleaf count but don't test negative cases (what if `graceBlocks = 0`? what if exercise is attempted after window?). Understandable for compiler integration tests, but edge-case tests would catch future regressions.

3. **`<tx.time >` trailing space** — Pre-existing compiler bug (also visible in `stability_offer.json`, `stability_vault.json`), not introduced by this PR. But worth a compiler fix.

---

### ✅ What's good

- Exercise/reclaim separation is clean — buyer-gated exercise, seller-gated reclaim
- Asset lookup via `OP_INSPECTOUTASSETLOOKUP` for stablecoin verification is correct
- Transfer functions correctly preserve all constructor params in continuation output
- Exit variants use `OP_CHECKSIGVERIFY` for intermediate sigs (correct N-of-N pattern)
- `OP_ADD64 + OP_VERIFY` guards against integer overflow on window computation
- options.md documentation is excellent — honest about limitations, clear on pre-sign protocol
- Test structure follows established patterns in the repo

---

**Requesting changes for P0 (transfer griefing vector) and P2 (graceBlocks minimum). P1 and P3 are compiler-level issues that should be filed as separate issues but shouldn't block this PR if documented.**

**🚨 Protocol-critical — requires human review before merge regardless of code changes.**

ghost (changes requested):
## Arkana Code Review — `CoveredCall` & `CashSecuredPut`

**⚠️ Protocol-critical: option contracts handling BTC and stablecoin settlement. Requires human sign-off.**

Good contract design overall — exercise is buyer-gated, reclaim is seller-gated, transfers preserve all terms, dual-path follows Arkade patterns correctly, overflow protection on `expiryHeight + graceBlocks` via `OP_ADD64 + OP_VERIFY`, and the docs (options.md) are thorough and honest about limitations. The concerns below are real and need to be addressed before merge.

---

### 🔴 P0 — Transfer functions lack exercise-window guards

**Files:** `examples/covered_call.ark:97-110`, `examples/cash_secured_put.ark:91-109`

`transferSeller` and `transferBuyer` have no timelock restriction. In the **cooperative path**, a seller can ask the Operator to co-sign a transfer during the exercise window, consuming the vault UTXO and **invalidating the buyer's pre-signed exercise template**.

The exit path is safe (requires buyer's signature). But the cooperative path relies entirely on Operator policy — nothing in the script prevents it.

**Impact:** Seller griefing vector. A seller who sees the option going deep ITM can transfer the vault to a fresh key, destroying the buyer's pre-signed exercise tx. The buyer has to re-negotiate exercise with the new "seller" (who is the same person with a new key).

**Recommendation:** Add `require(tx.time < expiryHeight, "transfers locked at expiry")` to both transfer functions. Yes, this has the compiler limitation noted in P1 — but even as documentation-in-code that the Operator must enforce, it's better than nothing. Alternatively, add a comment block explicitly stating the Operator MUST reject co-signing transfers at/after `expiryHeight`.

---

### 🟡 P1 — `tx.time < windowClose` compiles to non-functional ASM (compiler limitation)

**Files:** `examples/covered_call.ark:58`, `examples/cash_secured_put.ark:37`

The compiler grammar only handles `tx.time >= identifier` as a first-class pattern (→ `OP_CHECKLOCKTIMEVERIFY`). The `tx.time < windowClose` comparison falls through to generic `property_comparison`, emitting:

```
"<tx.time >",    ← note trailing space — parser bug
"<windowClose>",  ← local variable, unresolvable
"OP_LESSTHAN",
```

These are non-functional placeholders. In the cooperative path the Operator validates this off-chain, so the contract works. But the compiled script contains dead code that **cannot execute** and the upper-bound check exists only at the Operator trust level.

This is the first contract in the repo to use `tx.time <` — every other contract (htlc, swap, payment_auth, fuji_safe) only uses `tx.time >=`.

**Recommendation:** Either (a) file a compiler issue for `tx.time <` support, or (b) restructure to use only `tx.time >=` patterns (e.g., compute and pass `windowClose` as a constructor parameter, use `tx.time >= windowClose` as a "reclaim is valid" signal instead of "exercise is invalid"). At minimum, add a code comment explaining that this check is Operator-enforced only.

---

### 🟡 P2 — No minimum `graceBlocks` validation

**Files:** `examples/covered_call.ark:41`, `examples/cash_secured_put.ark:27`

If instantiated with `graceBlocks = 0`, the exercise window `[expiryHeight, expiryHeight + 0)` is empty — the buyer can **never** exercise. With `graceBlocks = 1`, the window is one block — practically impossible to hit.

Neither the contract nor the compiler warns about this. A misconfigured deployment silently creates an unexercisable option (seller keeps collateral + premium for free).

**Recommendation:** Add `require(graceBlocks >= 144, "grace period too short")` in the constructor or document a minimum in options.md. 144 blocks ≈ 1 day is the standard Arkade safety margin.

---

### 🟡 P3 — `<reclaimHeight>` placeholder in reclaim ASM is unresolvable

**Files:** `examples/covered_call.json` (reclaim variants), `examples/cash_secured_put.json` (reclaim variants)

The reclaim functions compute `reclaimHeight = expiryHeight + graceBlocks` as a local variable. The compiled ASM correctly computes this via `OP_ADD64` but then uses `<reclaimHeight>` as the `OP_CHECKLOCKTIMEVERIFY` argument:

```
"<expiryHeight>", OP_SCRIPTNUMTOLE64, "<graceBlocks>", OP_SCRIPTNUMTOLE64,
OP_ADD64, OP_VERIFY,
"<reclaimHeight>",      ← unresolvable placeholder
OP_CHECKLOCKTIMEVERIFY, OP_DROP,
```

The `OP_ADD64` result (the actual computed value) is still on the stack but isn't used by CLTV — instead, a separate unresolvable `<reclaimHeight>` placeholder is pushed. The SDK consuming this artifact would need to know to substitute `expiryHeight + graceBlocks` for `<reclaimHeight>`, but this isn't documented anywhere.

**Recommendation:** This is a compiler codegen issue — the CLTV should consume the ADD64 result from the stack rather than pushing a separate placeholder. File a compiler bug. For now, document in the JSON warnings or in options.md that `<reclaimHeight> = expiryHeight + graceBlocks`.

---

### 🟢 P4 — Nits

1. **JSON files missing trailing newline** — `covered_call.json:562`, `cash_secured_put.json:576`. Minor but causes noisy diffs.

2. **Test coverage is structural only** — Tests verify opcode presence and tapleaf count but don't test negative cases (what if `graceBlocks = 0`? what if exercise is attempted after window?). Understandable for compiler integration tests, but edge-case tests would catch future regressions.

3. **`<tx.time >` trailing space** — Pre-existing compiler bug (also visible in `stability_offer.json`, `stability_vault.json`), not introduced by this PR. But worth a compiler fix.

---

### ✅ What's good

- Exercise/reclaim separation is clean — buyer-gated exercise, seller-gated reclaim
- Asset lookup via `OP_INSPECTOUTASSETLOOKUP` for stablecoin verification is correct
- Transfer functions correctly preserve all constructor params in continuation output
- Exit variants use `OP_CHECKSIGVERIFY` for intermediate sigs (correct N-of-N pattern)
- `OP_ADD64 + OP_VERIFY` guards against integer overflow on window computation
- options.md documentation is excellent — honest about limitations, clear on pre-sign protocol
- Test structure follows established patterns in the repo

---

**Requesting changes for P0 (transfer griefing vector) and P2 (graceBlocks minimum). P1 and P3 are compiler-level issues that should be filed as separate issues but shouldn't block this PR if documented.**

**🚨 Protocol-critical — requires human review before merge regardless of code changes.**

ghost (changes requested):
## Arkana Code Review — Iterative (new commits pushed)

**⚠️ Protocol-critical: option contracts + compiler change affecting N-of-N exit leaves. Requires human sign-off.**

The rework is a significant improvement. Replacing exercise/reclaim/graceBlocks with a single permissionless `settle()` eliminates my previous P1 (tx.time < limitation), P2 (graceBlocks = 0), and P3 (reclaimHeight placeholder) entirely. The compiler change to filter oracle pubkeys from exit leaves is the right architectural call. My previous P0 (transfer griefing) is now lower severity since settle is permissionless, but not fully resolved — see below.

Focusing only on what's new since last review.

---

### Compiler change: `collect_data_only_pubkeys` — ✅ Correct

**File:** `src/compiler/mod.rs:152-358`

Verified all 33 `Expression` variants, all 6 `Requirement` variants, and all 5 `Statement` variants are handled in the new walker functions. The set-difference logic (`data_sigs - tx_sigs`) correctly identifies pubkeys that appear *only* in `checkSigFromStack` contexts. A pubkey used in both `checkSig` AND `checkSigFromStack` is correctly kept in the N-of-N — only pure data-signing keys are excluded.

The oracle key exclusion is critical: without it, unilateral exit requires Stork to co-sign L1 txs, which it can't do. This fix makes the exit path actually usable.

---

### 🟡 P1 — CashSecuredPut test file missing oracle exclusion test

**File:** `tests/cash_secured_put_test.rs`

`covered_call_test.rs` has `test_exit_leaf_excludes_oracle_pubkey` (line 110) — this is the key regression test for the compiler change. `cash_secured_put_test.rs` does NOT have this test. Both contracts exercise the same compiler path, but if the put contract ever diverges (different pubkey layout, additional oracle keys), the gap would be undetected.

**Recommendation:** Add the same test to `cash_secured_put_test.rs`.

---

### 🟡 P2 — Transfer functions still lack expiry guard (downgraded from previous P0)

**Files:** `examples/options/covered_call.ark:97-127`, `examples/options/cash_secured_put.ark:91-109`

My previous review flagged this as P0. With the new design, settle is permissionless (no buyer signature needed), so a post-expiry transfer can't block settlement — anyone can race the settle tx. This substantially reduces severity.

However, a cooperative-path transfer *after* expiry still consumes the vault UTXO. If a settle tx is already in the mempool and a transfer is mined first, the settle tx becomes invalid. The buyer has to re-construct settle against the new vault outpoint (which has a different seller key). Griefing, not theft, but still annoying.

**Recommendation:** Add a comment to both transfer functions: `// Operator policy: MUST reject co-signing transfers at or after expiryHeight`. This makes the trust assumption explicit in the contract source.

---

### 🟡 P3 — Playground description says "oracle-free" — factually wrong

**File:** `playground/main.js:18`

```js
description: 'European covered call + cash-secured put, physically settled, oracle-free',
```

These contracts are **oracle-triggered** — `settle()` requires a fresh oracle-signed price. "Oracle-free" is the opposite of what they are. This will confuse developers using the playground.

**Recommendation:** Change to `'European covered call + cash-secured put, physically settled, oracle-triggered'`.

---

### 🟢 P4 — Nits

1. **JSON files still missing trailing newline** — `covered_call.json`, `cash_secured_put.json`. Pre-existing from last review.

2. **Compiler warnings in JSON output** — The `<tx.time >` trailing space and unresolvable `<oracleAge>`/`<oracleMsg>` placeholders are pre-existing compiler issues, not introduced by this PR. But the new contracts are the first to surface the `<oracleAge>` / `<oracleMsg>` warnings (local variables used in the cooperative path). Worth filing as a compiler issue separately.

3. **No test for dual-usage pubkey** — A pubkey used in both `checkSig` and `checkSigFromStack` should NOT be excluded from the exit N-of-N. The logic is correct (`data_sigs.difference(&tx_sigs)`), but a unit test with a synthetic contract exercising this edge case would prevent regressions.

---

### ✅ What's improved since last review

- Exercise/reclaim complexity eliminated — single permissionless `settle()` is cleaner and removes 3 of my previous 4 findings
- Oracle signature verification is correct: `sha256(ticker + price + time)` with 144-block freshness window
- `collect_data_only_pubkeys` compiler change is complete and correct — all AST variants covered
- Exit leaves correctly produce 2-of-2 (seller+buyer) for settle, 3-of-3 (seller+buyer+newPk) for transfers — oracle key excluded
- options.md is rewritten and excellent — pre-signed OTM unwind fallback is well-documented with honest risk disclosure
- Test coverage for oracle exclusion (`test_exit_leaf_excludes_oracle_pubkey`) is the right regression test

---

**Requesting changes for P3 (factually wrong playground description) and P1 (missing test). P2 is a documentation request. None of these are blocking bugs — the core logic is correct.**

**🚨 Protocol-critical (compiler change affects ALL contracts with checkSigFromStack) — requires human review before merge.**

🤖 Reviewed by Arkana

ghost (commented):
## Arkana Code Review — Iterative (commit b4708449)

**⚠️ Protocol-critical fix. Requires human sign-off.**

---

### ✅ C1 fix is correct and necessary

`require(oracleTime >= expiryHeight, "oracle predates expiry")` — added to both `covered_call.ark:73` and `cash_secured_put.ark:62`.

**Reasoning verified:** The existing guards (`oracleAge ∈ [0, 144]` + `tx.time >= expiryHeight`) allowed a 144-block window where a pre-expiry oracle signature could be replayed. The commit message's repro scenario is accurate — buyer could use a stale ITM price from before expiry when the option is actually OTM at expiry. The fix constrains `oracleTime` to `[expiryHeight, tx.time]`, which is exactly right.

The compiled JSON output (`covered_call.json:155-157`, `cash_secured_put.json:155-157`) follows the same IR pattern as existing comparisons (`<oracleAge> OP_GREATERTHANOREQUAL 0`) — consistent with the compiler's intermediate representation.

---

### 🟡 Remaining items from previous review (still unaddressed)

1. **`test_settle_binds_oracle_time_to_expiry` missing from `cash_secured_put_test.rs`** — Same gap I flagged last time for `test_exit_leaf_excludes_oracle_pubkey`. The new test exists only in `covered_call_test.rs`. Both contracts exercise the same code path today, but if they diverge later, the gap is invisible. Low effort to add.

2. **`playground/main.js:19` still says "oracle-free"** — These contracts are oracle-triggered. Factually wrong.

3. **Transfer functions still lack expiry-guard comment** — Not blocking, but the cooperative-path griefing vector should be documented in-source.

---

### ✅ Test adequacy

`test_settle_binds_oracle_time_to_expiry` verifies `<expiryHeight>` is pushed ≥2 times in the settle ASM. This is a sufficient regression test — if the `require(oracleTime >= expiryHeight)` line is accidentally removed, the count drops to 1 and the test fails.

---

**No new blocking issues. The C1 fix is sound. Previous P1 and P3 from my last review remain open but are low-severity (missing test parity + wrong playground description). Approving the security fix itself.**

**🚨 Protocol-critical (oracle price replay prevention) — still requires human review before merge.**

🤖 Reviewed by Arkana

ghost (approved):
Approving — the C1 oracle-time-binding fix is correct and the regression test is adequate.

Remaining nits (missing test parity in cash_secured_put_test.rs, wrong playground description) are non-blocking.

🚨 Protocol-critical — human must still approve before merge.

🤖 Reviewed by Arkana

ghost (approved):
## Arkana Code Review — Iterative (commits b741dc7d, eb888623)

**C3 and C4 audit fixes reviewed. Both are correct.**

---

### ✅ C3 (`b741dc7d`) — strikePrice sanity check

`require(strikePrice > 0, "invalid strike price")` added to `settle()` in both `covered_call.ark:77` and `cash_secured_put.ark:64`. Compiled JSON confirms `<strikePrice> 0 OP_GREATERTHAN` emitted in the settle ASM for both contracts. Tests updated with matching inline source.

The int64 overflow rationale in `docs/options.md` and the `.ark` comment blocks is accurate: `strikePrice × btcSats` reaches ~1e19 for routine positions, exceeding int64. Doc-only + wallet enforcement is the correct tradeoff — both parties consent to the baked-in values by signing the funding tx.

`strikePrice > 0` prevents the degenerate case where every settle is trivially ITM. `oraclePrice > 0` was already present. Together they make the comparison `oraclePrice > strikePrice` meaningful in all cases. Cheap (3 bytes of script), no downside.

### ✅ C4 (`eb888623`) — +330 sat funding buffer

Doc-only. FUNDING BUFFER comments in both `.ark` files and a new subsection in `docs/options.md`. No logic changes — JSON diffs are timestamp-only recompilations.

The analysis is correct: every settle path produces one BTC-only output (`>= btcSats`) and one asset-bearing output (needs 330-sat P2TR dust carrier). Fund with exactly `btcSats` → vault is unspendable. Doc-only is the right call — lowering the contract check would change payout semantics.

---

### 🟡 Still open from previous reviews

1. **`playground/main.js:19`** — still says `"oracle-free"`. These contracts are oracle-triggered. Factually wrong. (P3, flagged 3 reviews ago)

2. **`test_settle_binds_oracle_time_to_expiry` missing from `cash_secured_put_test.rs`** — test parity gap. (P1, flagged 2 reviews ago)

Neither is blocking. The C3/C4 changes are sound.

**🚨 Protocol-critical — human must still approve before merge.**

🤖 Reviewed by Arkana

ghost (commented):
## Arkana Code Review — Iterative (commits 7b9c595, 72720b5, 4ccd3db, b412a2a)

**⚠️ Protocol-critical. Requires human sign-off.**

Reviewing 4 new commits since my last review: H5 (oracle freshness), M8 (transfer expiry guard), M9 (asset ID decomposition test), M10 (oracleTime documentation).

---

### ✅ H5 (`7b9c595`) — Oracle freshness tightened from 144 to 6 blocks

`require(oracleAge <= 6)` in both `covered_call.ark:98` and `cash_secured_put.ark:72`. Correct. Combined with C1's `oracleTime >= expiryHeight`, the settlement price is now constrained to `[expiryHeight, expiryHeight+6]` — a ~1h window vs the previous ~24h. This bounds the oracle MEV race where either party picks the most favorable signed price in the window.

6 blocks is a reasonable tradeoff: short enough to cap price drift, long enough for keepers to react. Stork signs every few seconds, so availability isn't an issue.

The comment in `covered_call.ark:91-96` explaining the rationale is good.

### ✅ M8 (`72720b5`) — Transfer functions guarded by expiry

`require(tx.time < expiryHeight, "no transfers after expiry")` added to both `transferSeller` and `transferBuyer` in both contracts. This was my P0/P2 from earlier reviews — now fully resolved on-chain, not just as operator policy.

Timeline is now clean and disjoint:
- `tx.time < expiryHeight` → transfers allowed, settle blocked
- `tx.time >= expiryHeight` → settle allowed, transfers blocked

Compiled JSON confirms `<tx.time > <expiryHeight> OP_LESSTHAN` emitted in cooperative transfer ASM for both contracts.

Regression test `test_transfers_guarded_by_expiry` in `covered_call_test.rs:248-263` verifies `<expiryHeight>` appears in both transfer functions' cooperative ASM. Adequate.

### ✅ M9 (`4ccd3db`) — Asset ID decomposition test

`test_asset_id_decomposes_to_txid_and_gidx` in `covered_call_test.rs:162-179` locks in the `bytes32 → (txid, gidx)` decomposition. Confirms `stableAssetId_txid` and `stableAssetId_gidx` are present, and raw `stableAssetId` is absent. Good regression test.

docs/options.md "Asset ID encoding" subsection is clear.

### ✅ M10 (`b412a2a`) — oracleTime documented as block height

Doc-only. Inline comments on `oracleTime` parameter in both `.ark` files explicitly state "BLOCK HEIGHT (Arkade nLockTime convention, NOT a Unix timestamp)". docs/options.md Oracle model section updated to match. Critical for SDK authors who would otherwise assume Unix seconds.

---

### 🟡 P1 — Stale "144-block" in covered_call.ark header comment

**File:** `examples/options/covered_call.ark:30`

```
// to feed (e.g. BTC/USD), price as attested value, timestamp + 144-block
// freshness window.
```

The code now uses 6 blocks (H5 fix), but the header comment still says 144. Misleading for anyone reading top-down.

**Fix:** Change to `6-block freshness window` in the header comment.

### 🟡 P2 — Test parity gap widening

**Files:** `tests/cash_secured_put_test.rs` vs `tests/covered_call_test.rs`

`covered_call_test.rs` now has **9** tests. `cash_secured_put_test.rs` has **5**. The put file is missing:
- `test_asset_id_decomposes_to_txid_and_gidx` (M9)
- `test_settle_binds_oracle_time_to_expiry` (C1)
- `test_exit_leaf_excludes_oracle_pubkey` (C2)
- `test_transfers_guarded_by_expiry` (M8)

I've flagged this every review. Both contracts exercise the same compiler paths *today*, but if they ever diverge, 4 regression tests silently don't cover the put contract. Copy-paste with `CALL_CODE → PUT_CODE` substitution. Low effort, high value.

### 🟡 P3 — `playground/main.js:19` still says "oracle-free"

```js
description: 'European covered call + cash-secured put, physically settled, oracle-free',
```

Flagged 4 reviews ago. These contracts are **oracle-triggered**. Wrong description in the playground.

---

### Status of previously flagged items

| Finding | Status |
|---|---|
| P0 transfer griefing (original review) | ✅ **Fixed** by M8 — on-chain expiry guard |
| P1 tx.time < limitation (original review) | ✅ **Fixed** by oracle-triggered redesign |
| P2 graceBlocks = 0 (original review) | ✅ **Fixed** by oracle-triggered redesign |
| P3 reclaimHeight placeholder (original review) | ✅ **Fixed** by oracle-triggered redesign |
| P1 test parity (2nd review) | 🟡 **Still open** — gap now 4 tests |
| P2 transfer expiry guard (2nd review) | ✅ **Fixed** by M8 |
| P3 playground "oracle-free" (2nd review) | 🟡 **Still open** |
| Stale header comment 144→6 | 🟡 **New** |

---

**All four new commits are correct and well-motivated. No blocking issues. The three 🟡 items are low-severity (stale comment, missing test parity, wrong playground text) and shouldn't block merge.**

**🚨 Protocol-critical (option contracts + compiler changes affecting exit leaves) — human must still approve before merge.**

🤖 Reviewed by Arkana

ghost (changes requested):
## Arkana Code Review — Iterative (commits 7b9c5953, 72720b5a)

H5 (oracle freshness tightening) and M8 (transfer expiry guard) reviewed.

---

### ✅ H5 (`7b9c5953`) — Tighten freshness window to 6 blocks

Sound. `oracleAge <= 6` replaces `<= 144` in both contracts. Combined with C1's `oracleTime >= expiryHeight`, the effective price window is `[expiryHeight, expiryHeight+6]` — roughly 1 hour, well within Stork's signing cadence. Applied consistently to `.ark`, `.json`, tests, and docs. No issues.

### 🔴 M8 (`72720b5a`) — Transfer expiry guard: **exit variants missing the guard**

The server (cooperative) variants are correct: `<tx.time> <expiryHeight> OP_LESSTHAN` emitted first in both `transferSeller` and `transferBuyer` ASM.

**But the exit (non-server) variants do not emit the guard.** Verified in both compiled JSONs:

```
# covered_call.json — transferSeller exit variant ASM starts with:
<sellerPk> <sellerPkSig> OP_CHECKSIGVERIFY <buyerPk> <buyerPkSig> ...
# ^^^ no <tx.time> <expiryHeight> OP_LESSTHAN anywhere

# Same for transferBuyer exit, and both transfer exit variants in cash_secured_put.json
```

The `.ark` source has `require(tx.time < expiryHeight)` in the function body unconditionally — the compiler should emit it in both variants. It doesn't. This is a **compiler code-generation bug**: the exit variant strips the `tx.time` introspection check.

**Severity assessment:** Medium. The exit variant requires N-of-N (seller + buyer) cooperation, so a *unilateral* post-expiry transfer is impossible via this path. But the M8 threat model includes collusion scenarios (e.g., buyer + compromised seller key transferring post-expiry to redirect a pending settle's payout). The fix is incomplete as long as the exit tapleaf doesn't enforce the guard.

**Likely root cause:** The compiler's exit-variant code generator probably skips `tx.time` introspection opcodes (which are Elements-specific). If this is by design, it needs to be documented and the M8 audit finding should note that the exit path relies on N-of-N cooperation as the sole guard. If it's a bug, the compiler needs to emit the check.

**Action:** Either fix the compiler to emit the guard in exit variants, or document in `docs/options.md` that the exit path's expiry safety relies solely on N-of-N cooperation.

### 🟡 Test gaps (M8)

1. **`test_transfers_guarded_by_expiry` only checks `server_variant == true`** (`covered_call_test.rs:228`). It would never catch the exit variant gap above. Should also assert on exit variants — or if the compiler intentionally strips the guard, the test should document why.

2. **`test_transfers_guarded_by_expiry` missing from `cash_secured_put_test.rs`** — Same pattern flagged in previous reviews for other tests.

### 🟡 Still open (flagged 3+ reviews ago)

3. **`playground/main.js:19`** — `description: 'European covered call + cash-secured put, physically settled, oracle-free'`. These contracts are oracle-triggered. "oracle-free" is factually wrong.

4. **`test_settle_binds_oracle_time_to_expiry` missing from `cash_secured_put_test.rs`** — Flagged 2 reviews ago.

---

**Summary:** H5 is clean. M8 server-variant fix is correct but the exit variant doesn't get the guard — this is either a compiler bug or a design limitation that needs documentation. Requesting changes for the exit-variant gap; the other items are non-blocking.

**🚨 Protocol-critical — human must still approve before merge.**

🤖 Reviewed by Arkana

ghost (commented):
## Arkana Code Review — Iterative (commits 197d398f, 056d8fa7)

Two minor fixes reviewed. Both correct, both resolve long-standing nits.

---

### ✅ `197d398f` — Playground description fix

`playground/main.js:19`: `oracle-free` → `oracle-triggered`. Correct. Resolves P3 flagged across 4 prior reviews.

### ✅ `056d8fa7` — Header comment freshness window

`examples/options/covered_call.ark:30`: `144-block` → `6-block`. Matches the H5 code change. JSON diff is timestamp-only recompilation. Correct.

---

### 🔴 Still open from previous review

**Exit-variant transfer guard missing** — `72720b5a` added `require(tx.time < expiryHeight)` to transfer functions, but the compiler only emits the check in server (cooperative) variants. Exit-variant ASM for `transferSeller` and `transferBuyer` in both contracts still lacks the `<tx.time> <expiryHeight> OP_LESSTHAN` guard. Either:
1. Fix the compiler to emit introspection checks in exit variants, or
2. Document that exit-path expiry safety relies solely on N-of-N cooperation

This was the blocking finding in my last review and remains unaddressed.

### 🟡 Test parity gap

`cash_secured_put_test.rs` still missing 4 tests present in `covered_call_test.rs`. Non-blocking but increasingly risky as the contracts accumulate invariants.

---

**No new issues in these two commits. Previous blocking issue (exit-variant guard) still open.**

**🚨 Protocol-critical — human must still approve before merge.**

🤖 Reviewed by Arkana

ghost (approved):
## Arkana Code Review — Iterative (commits 9045a891, 443a1e5c)

**Test parity + docs clarification. Both correct.**

### 9045a891 — `test(options): restore test parity between CashSecuredPut and CoveredCall`

Adds 4 CashSecuredPut tests mirroring existing CoveredCall coverage:
- `test_asset_id_decomposes_to_txid_and_gidx` (M9)
- `test_settle_binds_oracle_time_to_expiry` (C1)
- `test_exit_leaf_excludes_oracle_pubkey` (C2)
- `test_transfers_guarded_by_expiry` (M8)

All assertions are correct and consistent with the CoveredCall variants. Good that both contracts now have identical test coverage.

### 443a1e5c — `docs(options): clarify that the M8 transfer expiry guard is cooperative-only`

Documents the known limitation that `require(tx.time < expiryHeight)` on transfers is stripped from exit tapleaves (no introspection on exit paths — Arkade-wide constraint). The N-of-N consent argument is sound: a post-expiry transfer via the exit path requires both seller and buyer to sign, and neither party has an incentive to sign away their own settlement rights. No unilateral attack vector.

No contract code or compiler changes in these commits.

**⚠️ Protocol-critical: option contracts. Requires human sign-off before merge.**

🤖 Reviewed by Arkana

ghost (changes requested):
# 🔍 Arkana Code Review — arkade-os/compiler#33

**Reviewer:** Arkana (AI)  
**Verdict:** Request changes — **protocol-critical compiler change requires human sign-off**

---

## Executive Summary

The option contracts themselves (CoveredCall, CashSecuredPut) are well-designed and economically sound. The boundary logic is clean, the oracle model is solid, and the test coverage is good. **However, this PR contains a protocol-critical change to the compiler core (`collect_all_pubkeys` → `collect_data_only_pubkeys`) that changes exit-leaf N-of-N construction for ALL contracts, not just options.** This needs careful human review before merge.

---

## 🔴 PROTOCOL-CRITICAL — Compiler Core Change

### `src/compiler/mod.rs` — `collect_data_only_pubkeys` changes exit-leaf semantics globally

**What changed:** The old `collect_all_pubkeys` included ALL `pubkey`-typed constructor/function parameters in the N-of-N exit leaf. The new version walks the entire expression tree to classify pubkeys as transaction-signing (`checkSig`) vs data-signing (`checkSigFromStack`) and **excludes data-only pubkeys from the exit leaf**.

**This is the correct design** — oracle keys that only verify data signatures (Stork-style price attestations) should not be in the N-of-N because the oracle won't co-sign individual L1 transactions. The exit path would be unreachable otherwise.

**But the blast radius is large.** Every existing contract using `checkSigFromStack` gets a different exit-leaf N-of-N set when recompiled:

| Contract | oraclePk removed from exit N-of-N? | Old N-of-N | New N-of-N |
|---|---|---|---|
| **FujiSafe** | Yes | treasury + oracle | treasury only |
| **StabilityVault** | Yes | seeker + provider + oracle | seeker + provider |
| **StabilityOffer** | Yes | includes oracle | excludes oracle |
| **ThresholdOracle** | Yes | includes all oracle keys | excludes all oracle keys |
| **CoveredCall** (new) | N/A | N/A | seller + buyer |
| **CashSecuredPut** (new) | N/A | N/A | seller + buyer |

**This means recompiling existing contracts produces different script hashes.** Deployed VTXOs locked to the old scripts are unaffected (they remain valid), but any wallet or SDK that recompiles from source would get mismatched addresses.

⚠️ **Human must confirm:** Is this a deliberate breaking change for existing contracts, or should it be gated behind a compiler flag / version bump?

### Expression tree walker exhaustiveness: ✅ Verified

All **33** `Expression` variants, **6** `Requirement` variants, and **5** `Statement` variants have explicit match arms in the new walker functions. No wildcard `_` catch-all is used, so Rust's exhaustiveness checker will catch any future additions. This is correct and safe.

---

## 🟡 Medium Issues

### M1: PR description claims 8 tapleaves per contract — actual count is 6

The PR body says: *"each with 8 taproot leaves (4 functions × 2 variants)"*

The contracts define **3 functions** (settle, transferSeller, transferBuyer) × 2 variants = **6 tapleaves**. The tests correctly assert `out.functions.len() == 6`. The JSON artifacts also contain 6 entries each. The PR description is wrong.

### M2: `docs/options.md` also says 6 tapleaves per contract (line 125)

> Total tapleaves: **6** (CoveredCall) + **6** (CashSecuredPut) = 12.

This is consistent with the code. The PR description should be updated to match.

### M3: No defense-in-depth BTC value check on asset-bearing settle outputs

In every settle branch, the stablecoin-receiving output has an `assets.lookup()` check but no `value >= 330` check. Bitcoin's relay policy provides dust enforcement, so this isn't exploitable, but an explicit `require(tx.outputs[i].value >= 330, "dust carrier")` would be self-documenting and robust against future relay policy changes.

**Affected locations:**
- `covered_call.ark:113` — ITM output 0 (seller gets stablecoin, no value check)
- `covered_call.ark:135` — OTM output 1 (buyer gets stablecoin, no value check)
- `cash_secured_put.ark:91` — ITM output 1 (buyer gets stablecoin, no value check)
- `cash_secured_put.ark:103` — OTM output 0 (seller gets stablecoin, no value check)

Not a blocker, but worth adding.

### M4: No upper time bound on settlement past expiry

The contracts enforce `tx.time >= expiryHeight` but have no upper bound. Settlement at expiryHeight + 10,000 blocks (~10 weeks) is valid if the oracle provides a fresh attestation. This means the oracle has indefinite power to determine settlement price long after expiry. The pre-signed OTM unwind template with `STUCK_FUNDS_TIMELOCK` (documented in `docs/options.md`) mitigates this, but the SDK must actually implement it — there's no on-chain enforcement.

This is a design tradeoff documented in the docs, not a bug. Flagging for awareness.

---

## 🟢 Verified Correct

### Contract boundary logic

| Check | Result |
|---|---|
| ATM (oraclePrice == strikePrice) | Falls to OTM/unwind for both contracts ✅ |
| Transfer/settle race at expiryHeight | Mutually exclusive (`<` vs `>=`), no gap ✅ |
| Oracle freshness window | 6-block window, no permanent stuck state (oracle can re-attest) ✅ |
| Dust carrier accounting | btcSats + 330 - btcSats = 330, correct if settler is rational ✅ |
| Oracle replay protection | ticker + price + timestamp + 6-block freshness ✅ |
| Transfer collateral preservation | Both BTC value and stablecoin asset balance checked ✅ |
| Exit-leaf oraclePk exclusion | Correctly filtered by `collect_data_only_pubkeys` ✅ |
| Asset ID decomposition | `bytes32 stableAssetId` → `(stableAssetId_txid, stableAssetId_gidx)` verified in tests ✅ |

### Test coverage

16 integration tests across both contracts covering: compilation, oracle message reconstruction, exit-leaf pubkey exclusion, transfer expiry guards, collateral preservation, asset ID decomposition. Solid.

### Cross-repo impact

No downstream repos parse the compiler JSON output at runtime. The JSON schema is unchanged. SDKs (ts-sdk, go-sdk, rust-sdk, dotnet-sdk) build covenant scripts independently. No breakage from this PR.

---

## 📋 Summary of Requested Actions

1. **🔴 Human sign-off required** on the `collect_data_only_pubkeys` compiler change — it's correct but changes exit-leaf semantics for all existing contracts. Consider whether this needs a compiler version bump or migration note.
2. **🟡 Fix PR description** — 6 tapleaves per contract, not 8.
3. **🟡 Consider adding** explicit `value >= 330` checks on asset-bearing settle outputs (defense-in-depth).

Overall: high-quality PR with solid contract design. The contracts are economically sound and the oracle model is well-thought-out. The compiler change is the right fix. Just needs human eyes on the blast radius.

---
🤖 *Reviewed by Arkana — arkade-os/compiler code review agent*

PR #35, opened 2026-05-22
refactor(options): Rysk-faithful single-locked physical settlement

tiero:
## Why

The `CoveredCall` / `CashSecuredPut` shipped in #33 had both parties pre-lock their full exposure (seller's BTC *and* buyer's strike payment), settled deterministically by an oracle. 

Rysk's actual model has only the seller's collateral escrowed. The MM commits no capital until exercise — they only pay strike if the option is ITM and they choose to exercise. That capital efficiency is what makes the RFQ liquidity model work.

This PR replaces the dual-locked design with the faithful single-locked one.

### `CoveredCall`
| | Before (#33) | After (this PR) |
|---|---|---|
| Vault holds | btcSats BTC **+** strikeAmount stable | btcSats BTC **only** |
| Settlement | Oracle-triggered `settle()` deterministic ITM/OTM swap | Buyer's voluntary `exercise()` |
| Oracle | Required | None |
| Buyer's role at expiry | Passive (anyone broadcasts settle) | Must exercise if ITM, or forfeit |
| MM capital efficiency | Locks full strike value per option | Brings strike only at exercise |

### `CashSecuredPut`
Mirror — vault now holds stablecoin only; buyer brings BTC at exercise.

### Function shape (both)
- **`exercise(buyerSig)`** — valid from `expiryHeight`. Buyer pays the strike side, takes the underlying side.
- **`reclaim(sellerSig)`** — valid from `expiryHeight + graceBlocks`. Seller's path when buyer didn't exercise.
- **`transferSeller`** / **`transferBuyer`** — pure key swap, pre-expiry only. Each preserves the appropriate collateral (BTC value for the call vault, stablecoin asset balance for the put vault).

8 tapleaves per contract (4 functions × cooperative/exit). 9 tests per contract, full parity between the two.

## Worked example (from the quant)

1 BTC notional, $90k strike, $1,500 premium upfront. Expiry Jun 26.

| Spot at expiry | Exercise? | Seller ends with | Buyer ends with |
|---|---|---|---|
| $80k (OTM) | No | 1 BTC + $1,500 premium | nothing besides lost premium |
| $120k (ITM, exercised) | Yes | $90k USDT + premium | 1 BTC (now worth $120k) — $28.5k PnL net of premium |
| $120k (ITM, **buyer ghosts**) | No | 1 BTC + premium (free upside) | −premium (forfeited ITM gain) |

The last row is the seller's "Christmas morning" outcome the quant mentioned — *good for the seller if ITM and buyer does not exercise*. The buyer's only protection against this is broadcasting their exercise tx within the grace window.

## What this gains and trades away

**Gains:**
- Capital efficiency for the MM (no pre-locked strike across the full option lifetime). Matches Rysk's RFQ liquidity model.
- No oracle dependency — the buyer's exercise decision *is* the settlement signal.
- Simpler ASM: no oracle message reconstruction, no ITM/OTM branching in script.

**Trades away:**
- Buyer liveness requirement: buyer must broadcast within the grace window or forfeit any ITM gain plus the premium.
- Operator-down asymmetry: an ITM buyer cannot force exercise unilaterally if the Operator is offline AND the seller refuses to co-sign the N-of-N exit. Pre-signed exit-path templates at funding time mitigate but don't fully solve. Documented in `docs/options.md`.

The previous dual-locked variant traded capital efficiency for liveness independence. Different product, different audience.

## Test plan

- [x] `cargo test` — all 28 suites pass, including 9 new tests per contract
- [x] `cargo fmt --check` clean
- [x] Both `.ark` files compile cleanly via `arkadec`
- [x] Playground regenerated (`playground/generate_contracts.sh`)
- [x] Manual review of compiled ASM: exercise uses CLTV + checksig + asset lookup + value check; reclaim is pure timelock + checksig; transfers are key-swap with the appropriate collateral check; exit leaves carry no introspection

## Known compiler limitations (called out in the doc)

1. `tx.time < expiryHeight` in transfers compiles to a non-functional placeholder on the exit variant — the cooperative path Operator validates this off-chain, and the exit path's N-of-N consent removes the unilateral griefing vector.
2. `reclaimHeight = expiryHeight + graceBlocks` is computed via `OP_ADD64 + OP_VERIFY` but then a separate `<reclaimHeight>` placeholder is pushed for CLTV. The SDK substitutes the computed value when building the witness. Same pattern Arkana flagged on PR #33.

Both are upstream compiler issues, not contract bugs.

https://claude.ai/code/session_01MXpCrecXFzQ3bmHqUuW5yx

---
_Generated by [Claude Code](https://claude.ai/code/session_01MXpCrecXFzQ3bmHqUuW5yx)_

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -1,6 +1,6 @@
 # Options — How It Works
 
-**Status:** v1 design complete
+**Status:** v2 — single-locked Rysk-faithful design (supersedes the earlier dual-locked variant in PR #33)
 
 ---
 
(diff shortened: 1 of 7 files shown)


ghost (changes requested):
## 🔒 Protocol-Critical Review — Requesting Human Sign-Off

This PR replaces the dual-locked, oracle-triggered settlement model (PR #33 + audit fixes C1/C3/C4/H5/M8/M9/M10) with a single-locked, buyer-exercised model. The design is sound and well-documented. No cross-repo consumers exist — changes are contained to the compiler repo.

I'm requesting changes **not because the code is wrong**, but because this is protocol-critical (settlement logic, collateral handling, exit paths) and per policy requires explicit human approval.

---

### ✅ What's correct

1. **Exercise output constraints are tight.** Both contracts properly verify output scriptPubKey AND value/asset-balance on both outputs 0 and 1. Seller gets paid, buyer gets the underlying. No way to redirect funds.

2. **Reclaim is correctly unconstrained on outputs.** The seller's reclaim only gates on timelock + sellerSig. Since it's the seller's own collateral, no output constraints are needed. This is right.

3. **Transfer collateral checks correctly match the vault contents.** CoveredCall transfers check BTC value only (vault is BTC-only). CashSecuredPut transfers check stablecoin asset balance only (vault is stablecoin-only). The old dual-check for both legs is correctly removed.

4. **Exit-path N-of-N is correct.** Exercise exit = seller+buyer. Reclaim exit = seller only. Transfers exit = seller+buyer+new party. CLTV is correctly preserved on the reclaim exit path (it's pure Bitcoin script, not introspection).

5. **Oracle removal is clean.** No residual oracle references in any function, any variant. Confirmed by tests.

6. **Tests are comprehensive.** 9 tests per contract, covering: leaf count, signature requirements, no-oracle invariant, output verification, reclaim timelock, asset ID decomposition, transfer expiry guard, collateral preservation, exit-path purity.

---

### ⚠️ Findings

**P1 (Medium) — `exercise()` has no upper time bound; grace window is soft**
`covered_call.ark:55`, `cash_secured_put.ark:51`

```
function exercise(signature buyerSig) {
    require(tx.time >= expiryHeight, "before expiry");  // ← no upper bound
```

`exercise()` is valid from `expiryHeight` **forever**. The `graceBlocks` parameter only controls when `reclaim()` becomes valid — it doesn't *disable* exercise. After `expiryHeight + graceBlocks`, both exercise and reclaim are valid simultaneously, and it's a mempool race.

The docs acknowledge this ("Race between seller's reclaim and any belated buyer exercise"), so it's intentional. But the term "grace window" implies a hard window when it's actually soft. A seller who is slow to broadcast reclaim at `reclaimHeight` could lose the race to a belated exercise.

**Suggestion:** Either:
- Add a comment to both `.ark` files clarifying this is a race, not a hard window, OR
- Consider whether `require(tx.time < expiryHeight + graceBlocks)` should be added to exercise for a hard cutoff. The tradeoff: a hard cutoff means if the cooperative path is down and the seller is unresponsive, a buyer who's 1 block late loses their ITM gain with no recourse. The soft race is arguably more forgiving.

Not blocking — this is a design choice, not a bug.

---

**P2 (Low) — No minimum `graceBlocks` enforced on-chain**
`covered_call.ark:60`, `cash_secured_put.ark:56`

If `graceBlocks = 0`, then `reclaimHeight = expiryHeight`. The seller's reclaim becomes valid at the exact same block exercise opens. The buyer's exercise window is zero-length in practice (the seller will race them from block 1).

The contract doesn't prevent this. SDK/wallet must enforce a sane minimum (e.g., ≥ 6 blocks / ~1 hour). Worth adding to `docs/options.md` under risk disclosures or as an inline comment.

---

**P3 (Info) — `OP_ADD64` result is left on the stack in `reclaim` ASM**
`covered_call.json` reclaim cooperative ASM, `cash_secured_put.json` reclaim cooperative ASM

```
<expiryHeight> OP_SCRIPTNUMTOLE64
<graceBlocks>  OP_SCRIPTNUMTOLE64
OP_ADD64       ← pushes (sum, success_flag)
OP_VERIFY      ← pops flag, leaves sum on stack
<reclaimHeight> OP_CHECKLOCKTIMEVERIFY OP_DROP
```

The computed sum from `OP_ADD64` stays on the stack but is never compared to `<reclaimHeight>`. The SDK can push any value as `<reclaimHeight>` and the script won't catch a mismatch.

This is **safe** because:
- The tapscript (including the baked-in `<reclaimHeight>`) is committed at funding time
- Both parties inspect and sign the funding tx
- Nobody can change the placeholder post-funding
- The OP_ADD64 serves as overflow protection only

This is the known compiler limitation documented in the PR. Flagging for completeness.

---

**P4 (Info) — Stack cleanliness after `reclaim` cooperative path**
Same ASM as P3. After execution:
- `OP_ADD64` result (8 bytes) remains on the stack
- Final `OP_CHECKSIG` for serverSig leaves TRUE/FALSE

The ADD64 leftover could cause a dirty stack. Whether this matters depends on the Arkade script execution rules (Elements/Liquid `SCRIPT_VERIFY_CLEANSTACK`). If cleanstack is enforced, this is a bug. If the compiler wraps scripts to handle this, it's fine. Worth a quick check.

---

**P5 (Info) — Compiled JSON warnings about `<reclaimHeight>` placeholder**
`covered_call.json`, `cash_secured_put.json`

Both compiled JSONs emit:
```
warning[output-invariant]: fn 'reclaim' (serverVariant=true): placeholder <reclaimHeight> is not in witnessSchema or constructorInputs
warning[output-invariant]: fn 'reclaim' (serverVariant=false): placeholder <reclaimHeight> is not in witnessSchema or constructorInputs
```

SDK authors must know to compute `reclaimHeight = expiryHeight + graceBlocks` and substitute. This is documented, but the warnings are a breadcrumb for anyone integrating.

---

### 📋 Test coverage gaps (minor)

1. **No test for `graceBlocks = 0` edge case.** A test asserting the compiled reclaim ASM pushes both `<expiryHeight>` and `<graceBlocks>` (confirming the ADD64 is present) would lock in overflow protection.

2. **No test asserting reclaim exit-path has CLTV.** The existing `test_exit_leaves_have_no_introspection` exempts reclaim from the CLTV check (`if fn_name != "reclaim"`), but doesn't positively assert that reclaim's exit path DOES have CLTV. A one-liner would lock this in.

---

### 🏗️ Cross-repo impact

**None.** No SDK, demo, or downstream repo references CoveredCall or CashSecuredPut. These contracts are compiler-internal examples. The API break (settle→exercise, removed oracle params, added graceBlocks) has zero external blast radius today.

---

### Bottom line

Clean refactor. The design tradeoffs (buyer liveness, soft grace window, operator-down asymmetry) are well-articulated in docs and the PR description. The quant's Rysk model is faithfully implemented. Code quality is high.

Requesting human sign-off because this is protocol-critical settlement logic. P1 and P2 are worth a human opinion on whether the soft grace window and missing graceBlocks minimum are acceptable product decisions.

🤖 Reviewed by Arkana

PR #41, opened 2026-06-10
feat: InventoryHedge — market-maker inventory-hedging vault

tiero:
## What

A new `.ark` contract — **`examples/hedging/inventory_hedge.ark`** — that sketches the on-chain core of the design note `docs/mm-residual-hedge.md`. It is a perpetual, BTC-settled, fully-collateralized market-maker inventory hedge that generalizes `examples/stability/stability_vault.ark`:

- **claim leg** (`claimPk`, the desk): holds a fiat claim `targetFiat`, delta-flat.
- **long leg** (`longPk`, the treasury): holds the residual BTC upside, posts the over-collateral, earns funding.
- Oracle-marked (Fuji pattern), no liquidation (settlement is a clamp into `[0, totalCollateral]`), one instance per `BTC/<fiat>` pair via `ticker`.

### Functions (6 → 12 tapleaves)
`transfer`, `updateFunding`, `addCapital`, `removeCapital`, `redeem` (desk redeems its fiat claim for BTC), `withdraw` (treasury-driven settlement).
Funding is rolled on-chain (`>= 0` guard + accrual); the dynamic, imbalance-driven *value* is computed off-chain by the desk's risk engine (design note §4) and supplied to `updateFunding`.

### Also in this PR
- `tests/inventory_hedge_test.rs` — roundtrip (both variants emit, non-empty witness schemas), oracle-path checks (`OP_CHECKSIGFROMSTACK` + `OP_DIV64`), `redeem` clamp branches, `<SERVER_KEY>` injection, and `updatedAt`-stripped determinism. **8 tests, green.**
- `docs/mm-residual-hedge.md` — design note **plus** "Part II — Transaction flows" (per-function input/output layouts), now a single consolidated doc.
- `playground/main.js` — **Hedging** project folder wiring `inventory_hedge.ark` into the WASM playground. (`playground/contracts.js` is gitignored and regenerated by `./playground/generate_contracts.sh`.)

## Acceptance
- `cargo run -- examples/hedging/inventory_hedge.ark -o /tmp/ih.json` ✅ compiles; only the same benign `output-invariant` warnings `stability_vault.ark` emits (computed spend-time locals used in `value >= X` checks).
- `cargo fmt --check` ✅ clean.
- `cargo test` ✅ green (full suite, including the 8 new tests).

## Dialectical review
A self-review in thesis / antithesis / synthesis is in the first comment, including an empirical comparison against an alternative one-shot draft (compiled in this toolchain: its `min()` clamp, `new P2TR(pk)` payouts, and `bytes`-bound state script all degrade to unbound placeholders — the clamp, payout destination, and state recursion go unenforced). Open questions for reviewers are flagged there.

> Draft: opened for review, not for merge.

## Summary by CodeRabbit

* **New Features**
  * Added a perpetual BTC-settled, fully-collateralized inventory hedge contract for market-making desks.
  * Added a playable Playground project demonstrating the hedge contract.

* **Documentation**
  * Added comprehensive design and lifecycle documentation for the hedging mechanism.

* **Tests**
  * Added test suite validating compilation, ABI/ASM invariants, oracle-settlement behavior, and determinism.


tiero:
## Dialectical review

### Thesis — the case *for* this implementation

- **Faithful to the reference.** It mirrors `stability_vault.ark`'s proven shapes: self-replacement state commitment via `require(tx.outputs[0].scriptPubKey == new InventoryHedge(...))`, the Fuji oracle (`checkSigFromStack(sig, oraclePk, sha256(ticker + price + time))`), the interleaved `/1e6` funding accrual, the `>= 0` clock-regression guards, and `new SingleSig(pk)` payout targets. Nothing exotic — every construct is already load-bearing in a shipped contract.
- **It actually compiles to enforceable script.** 12 tapleaves (6 fns × 2 variants), both variants carry non-empty witness schemas, `<SERVER_KEY>` is auto-injected on every cooperative leaf and never a constructor param, and the `claimExit` clamp is real (three `OP_IF` branches + the 330-sat dust guard). The only `output-invariant` warnings are the **same benign class** `stability_vault.ark` emits — computed locals (`delta`, `claimRaw`, `longPayout`) used inside `value >= X` comparisons, not spender-supplied placeholders.
- **Scope discipline.** Per the spec, funding stays a `>= 0` guard + accrual; the imbalance-driven *value* lives off-chain (design note §4). `removeCapital` computes accrual only for the ratio guard and deliberately does **not** roll `lastUpdate`, so a withdrawal never truncates funding owed the desk.
- **Gates pass:** `cargo run … -o /tmp/ih.json`, `cargo fmt --check`, and `cargo test` (full suite + 8 new tests) are all green.

### Antithesis — the case *against* (and an alternative draft)

A separate one-shot draft (`min()` clamp, `new P2TR(pk)` payouts, `bytes nextScript = new InventoryHedge(...)`) reads *cleaner* than this PR. So is terse-but-untested better? I compiled it in this toolchain to find out — and the readability hides three correctness holes the compiler exposes:

1. **`min(claimRaw, totalCollateral)` is not a builtin.** It compiles to a literal placeholder token `<min(claimRaw, totalCollateral)>` — an unbound value the spender supplies freely. **The clamp is not enforced on-chain.** This PR uses nested `if`/`require` (uglier under single-assignment, but the bound is real).
2. **`new P2TR(pk)` references a contract that doesn't exist in the repo** and the draft never imports `single_sig.ark`; it emits `<VTXO:P2TR(...)>` pointing at nothing. The repo convention is `new SingleSig(pk)`.
3. **`bytes nextScript = new InventoryHedge(...)` then compare** breaks VTXO-placeholder resolution: it emits `<nextScript>` ("cannot be constructed without a binding") instead of inlining `<VTXO:InventoryHedge(...)>`. **State recursion isn't actually enforced.** The working examples always inline `new Contract(...)` directly in the `require`.

So "compiles successfully" ≠ "correct": the draft compiles *and* leaves the clamp, the payout destination, and the state transition all unbound. The lesson cuts at this PR too — a green `cargo run` is necessary, not sufficient; the behavioral asserts in the test file are what make the difference legible.

But the draft lands two fair hits worth keeping:
- **int64 overflow.** `newTargetFiat * 1e8` overflows int64 around `targetFiat ≈ 9.2e10` minor units. This PR inherits that bound from `stability_vault.ark` (only funding is interleaved). For plausible single-vault sizes it's fine; for very large claims the sats conversion would need interleaving too.
- **min() *would* be cleaner** if it existed — the nested-`if` clamp is verbose. A future compiler builtin would simplify three functions here.

### Synthesis

Prefer the implementation whose guarantees are **machine-checked against this compiler**, not the one that reads best. This PR is the conservative generalization: same primitives as the reference, clamp/recursion/payout all genuinely enforced, gates green. Fold in the draft's *insights* (document the int64 ceiling; consider a `min`/`clamp` builtin later) without its *constructs* (`min`, `P2TR`, bytes-bound state), which don't survive contact with the toolchain.

**Open questions for reviewers:**
1. **Overflow ceiling** — is `targetFiat` bounded well under ~9.2e10 minor units in practice, or should `claimRaw` interleave like the funding math?
2. **`merge`/`split`** — the design note §6 lists them for resizing; the spec scoped them out. Add as a fast-follow, or keep this minimal?
3. **`longExit` symmetry** — should a provider-driven exit charge the desk anything, or stay fee-free as here?

*Draft — opened for review, not merge.*

---
_Generated by [Claude Code](https://claude.ai/code)_

ghost (changes requested):
## 🔍 Arkana Code Review — `InventoryHedge`

**⚠️ PROTOCOL-CRITICAL: This PR adds a new settlement contract with oracle-priced exits, collateral ratio enforcement, and funding accrual. Requires human sign-off regardless of code quality.**

Overall this is a clean generalization of `StabilityVault` with the exit fee removed and naming adapted for the hedging use case. The core settlement math, oracle pattern, anti-grief rule, and clamp logic are correctly carried over. The rename to `redeem`/`withdraw` in the latest commit is clean. No downstream consumers exist yet.

I found **one substantive issue** and a few items worth discussing before this leaves draft.

---

### 🔴 Issue — `removeCapital` missing `newTargetFiat > 0` guard

**`inventory_hedge.ark:116`** (the `removeCapital` function)

`updateFunding` correctly guards `require(newTargetFiat > 0, "claim wiped by funding")`, but `removeCapital` computes the same `newTargetFiat = targetFiat + delta` and feeds it into the collateral-ratio check **without** validating it's positive:

```ark
int newTargetFiat      = targetFiat + delta;
int currentClaimBase   = newTargetFiat * 100000000 / oraclePrice;
int minCollateral      = currentClaimBase * (100 + collateralRatioPct) / 100;
```

If the initial `fundingRatePerSec` is negative (which the constructor does not prevent — the `>= 0` constraint only applies to `updateFunding`'s *new* rate), and enough time passes before the first `updateFunding` call, `newTargetFiat` can go negative → `currentClaimBase` negative → `minCollateral` negative → the ratio check `newTotalCollateral >= minCollateral` becomes trivially true → the long leg drains the vault.

Yes, this is the same pattern as StabilityVault. And yes, in practice the initial rate should be non-negative for a hedging vault. But defense-in-depth says: **add `require(newTargetFiat > 0, "claim wiped by funding")` in `removeCapital`** (and arguably in `redeem`/`withdraw` too, before the clamp — though the clamp handles it safely, an explicit guard documents intent).

Same applies to `redeem` and `withdraw` at lines ~651 and ~696. The clamp (`if (claimRaw <= 0)`) handles the case correctly, but an explicit `require(newTargetFiat > 0)` before the conversion makes the invariant visible and prevents subtle bugs if the clamp logic is ever refactored.

---

### 🟡 Observations (non-blocking, but worth considering)

**1. Doc comment says "signed" but contract enforces `>= 0`** — `inventory_hedge.ark:10` says `fundingRatePerSec` is a "signed fixed-point fraction" but `updateFunding` enforces `newFundingRatePerSec >= 0`. The *initial* rate at construction is the only place a negative value could enter. If negative initial rates are not intended for the hedging use case, consider calling this out explicitly in the doc (or adding a construction-time guard if the language supports it). Current wording could mislead an integrator into thinking negative rates are supported at runtime.

**2. Doc §6 references `split` for resizing but the contract omits it** — The design doc talks about "split → peel off part of the claim when D shrinks" and "merge → combine positions when D grows." These are deferred to "thinnest first build" per §10, but the doc still reads as if they exist. Consider either (a) noting in the doc that split/merge are deferred, or (b) adding a forward-reference to a future PR.

**3. Compiler warnings in the JSON artifact** — The compiled `inventory_hedge.json` carries 50+ `output-invariant` warnings (e.g., `placeholder <tx.offchainTime> is not in witnessSchema`). These are expected for computed intermediates and match StabilityVault's pattern, but shipping them in the committed JSON is noisy. If the compiler has a way to suppress expected warnings, it'd clean up the artifact.

**4. Test coverage is compilation-level only** — The test suite (`tests/inventory_hedge_test.rs`) pins structural properties (12 tapleaves, SERVER_KEY injection, OP_IF branches for the clamp). There are no behavioral/simulation tests for edge cases like: anti-grief rule triggering, collateral ratio boundary, dust threshold routing, zero-elapsed funding. This matches StabilityVault's test pattern, but for a protocol-critical contract, consider adding property-based or scenario tests that exercise the arithmetic with concrete values.

---

### ✅ What looks good

- Oracle verification: Fuji-pattern `sha256(ticker + price + time)` + `checkSigFromStack` + 600s freshness window — correct and consistent with StabilityVault.
- Anti-grief rule in `updateFunding`: correctly prevents zero-accrual clock advancement when `fundingRatePerSec != 0`.
- Settlement clamp in `redeem`/`withdraw`: three-way branch (claim wiped / fully covered / split) with 330-sat dust guard — correct.
- Self-replacement pattern: all non-terminal functions correctly assert `tx.outputs[0].scriptPubKey == new InventoryHedge(...)` with the next state.
- Exit variants: 2-of-2 (or 3-of-3 for transfer) multisig + CLTV — standard operator-liveness backstop.
- `removeCapital` correctly uses accrued funding for the ratio check without rolling it into the vault state (no clock advancement).
- Playground integration is clean — just a new project entry.
- Rename to `redeem`/`withdraw` (commit 3) was applied consistently across .ark, .json, .md, and .rs files.

---

**Verdict:** Request changes on the `newTargetFiat > 0` defense-in-depth guard. Everything else is solid. **Flagging for mandatory human review** — this is a new settlement contract that handles real funds.

ghost (changes requested):
## 🔍 Arkana Code Review — `InventoryHedge`

**⚠️ PROTOCOL-CRITICAL: This PR adds a new settlement contract with oracle-priced exits, collateral ratio enforcement, and funding accrual. Requires human sign-off regardless of code quality.**

Overall this is a clean generalization of `StabilityVault` with the exit fee removed and naming adapted for the hedging use case. The core settlement math, oracle pattern, anti-grief rule, and clamp logic are correctly carried over. The rename to `redeem`/`withdraw` in the latest commit is clean. No downstream consumers exist yet.

I found **one substantive issue** and a few items worth discussing before this leaves draft.

---

### 🔴 Issue — \`removeCapital\` missing \`newTargetFiat > 0\` guard

**\`inventory_hedge.ark:116\`** (the \`removeCapital\` function)

\`updateFunding\` correctly guards \`require(newTargetFiat > 0, \"claim wiped by funding\")\`, but \`removeCapital\` computes the same \`newTargetFiat = targetFiat + delta\` and feeds it into the collateral-ratio check **without** validating it's positive:

\`\`\`ark
int newTargetFiat      = targetFiat + delta;
int currentClaimBase   = newTargetFiat * 100000000 / oraclePrice;
int minCollateral      = currentClaimBase * (100 + collateralRatioPct) / 100;
\`\`\`

If the initial \`fundingRatePerSec\` is negative (which the constructor does not prevent — the \`>= 0\` constraint only applies to \`updateFunding\`'s *new* rate), and enough time passes before the first \`updateFunding\` call, \`newTargetFiat\` can go negative → \`currentClaimBase\` negative → \`minCollateral\` negative → the ratio check \`newTotalCollateral >= minCollateral\` becomes trivially true → the long leg drains the vault.

Yes, this is the same pattern as StabilityVault. And yes, in practice the initial rate should be non-negative for a hedging vault. But defense-in-depth says: **add \`require(newTargetFiat > 0, \"claim wiped by funding\")\` in \`removeCapital\`** (and arguably in \`redeem\`/\`withdraw\` too, before the clamp — though the clamp handles it safely, an explicit guard documents intent).

Same applies to \`redeem\` and \`withdraw\` at lines ~651 and ~696. The clamp (\`if (claimRaw <= 0)\`) handles the case correctly, but an explicit \`require(newTargetFiat > 0)\` before the conversion makes the invariant visible and prevents subtle bugs if the clamp logic is ever refactored.

---

### 🟡 Observations (non-blocking, but worth considering)

**1. Doc comment says "signed" but contract enforces \`>= 0\`** — \`inventory_hedge.ark:10\` says \`fundingRatePerSec\` is a "signed fixed-point fraction" but \`updateFunding\` enforces \`newFundingRatePerSec >= 0\`. The *initial* rate at construction is the only place a negative value could enter. If negative initial rates are not intended for the hedging use case, consider calling this out explicitly in the doc (or adding a construction-time guard if the language supports it). Current wording could mislead an integrator into thinking negative rates are supported at runtime.

**2. Doc §6 references \`split\` for resizing but the contract omits it** — The design doc talks about "split → peel off part of the claim when D shrinks" and "merge → combine positions when D grows." These are deferred to "thinnest first build" per §10, but the doc still reads as if they exist. Consider either (a) noting in the doc that split/merge are deferred, or (b) adding a forward-reference to a future PR.

**3. Compiler warnings in the JSON artifact** — The compiled \`inventory_hedge.json\` carries 50+ \`output-invariant\` warnings (e.g., \`placeholder <tx.offchainTime> is not in witnessSchema\`). These are expected for computed intermediates and match StabilityVault's pattern, but shipping them in the committed JSON is noisy. If the compiler has a way to suppress expected warnings, it'd clean up the artifact.

**4. Test coverage is compilation-level only** — The test suite (\`tests/inventory_hedge_test.rs\`) pins structural properties (12 tapleaves, SERVER_KEY injection, OP_IF branches for the clamp). There are no behavioral/simulation tests for edge cases like: anti-grief rule triggering, collateral ratio boundary, dust threshold routing, zero-elapsed funding. This matches StabilityVault's test pattern, but for a protocol-critical contract, consider adding property-based or scenario tests that exercise the arithmetic with concrete values.

---

### ✅ What looks good

- Oracle verification: Fuji-pattern \`sha256(ticker + price + time)\` + \`checkSigFromStack\` + 600s freshness window — correct and consistent with StabilityVault.
- Anti-grief rule in \`updateFunding\`: correctly prevents zero-accrual clock advancement when \`fundingRatePerSec != 0\`.
- Settlement clamp in \`redeem\`/\`withdraw\`: three-way branch (claim wiped / fully covered / split) with 330-sat dust guard — correct.
- Self-replacement pattern: all non-terminal functions correctly assert \`tx.outputs[0].scriptPubKey == new InventoryHedge(...)\` with the next state.
- Exit variants: 2-of-2 (or 3-of-3 for transfer) multisig + CLTV — standard operator-liveness backstop.
- \`removeCapital\` correctly uses accrued funding for the ratio check without rolling it into the vault state (no clock advancement).
- Playground integration is clean.
- Rename to \`redeem\`/\`withdraw\` (commit 3) was applied consistently across .ark, .json, .md, and .rs files.

---

**Verdict:** Request changes on the \`newTargetFiat > 0\` defense-in-depth guard. Everything else is solid. **Flagging for mandatory human review** — this is a new settlement contract that handles real funds.

tiero:
Thanks — substantive review. Addressed in `9ed13de`.

### 🔴 `removeCapital` guard — fixed
Agreed, this is the real one. Added `require(newTargetFiat > 0, "claim wiped by funding")` in `removeCapital` before the ratio math, matching `updateFunding`. Without it a negative *initial* `fundingRatePerSec` (not prevented at construction) could drive `newTargetFiat`→ `minCollateral` ≤ 0 and trivially pass the ratio check, draining the vault. `"claim wiped by funding"` now appears in both `updateFunding` and `removeCapital`.

### `redeem` / `withdraw` — intentionally **not** guarded (with reasoning)
I did **not** add `require(newTargetFiat > 0)` here, and I'd argue it would be a regression rather than defense-in-depth. These are *terminal* settlements, and the `claimRaw <= 0` branch is the **correct** outcome for a wiped claim: route the entire collateral to the long leg. A hard `require` before the clamp would **abort a legitimate settlement** — including the long leg's own `withdraw`, which must succeed precisely when the claim is worthless so the treasury can recover its collateral. The clamp is the invariant, not a `require`. I added an inline comment to both functions documenting this on purpose so a future refactor doesn't "fix" it into a bug.

The distinction vs `removeCapital`: `removeCapital` is *non-terminal* (keep the vault alive, reclaim only excess) — there a wiped claim has no valid semantics and should force a `withdraw` instead, so aborting is right.

### 🟡 Observations
1. **"signed" doc wording** — clarified. The constructor comment now states runtime updates enforce `>= 0` and a negative value can only enter at construction (and shouldn't, for a hedging vault).
2. **`split`/`merge` deferred** — added an explicit "Deferred (fast-follow)" note in the design doc's Part II: they're reusable from `stability_vault.ark` but out of scope for the §10 thinnest build; only `transfer` ships now.
3. **JSON warnings** — these are the same benign `output-invariant` class `stability_vault.ark` emits (computed spend-time locals used in `value >= X` checks, never spender-supplied). Suppressing them would be a change to the gated compiler (`src/compiler`), out of scope for this contract PR; happy to file a separate issue for an "expected-intermediate" annotation.
4. **Behavioral tests** — worth flagging that this repo has **no script interpreter**; it's a compiler, so every contract's tests (StabilityVault included) are necessarily compilation/ASM-structural, not arithmetic simulation. Property/scenario execution would need a Tapscript evaluator that doesn't exist here. The new test pins the structural invariants (12 tapleaves, `<SERVER_KEY>`, the clamp's `OP_IF` branches, oracle-path opcodes); concrete-value simulation is a separate harness, not something I can add in-tree today.

Staying in **draft** — acknowledging the protocol-critical flag; this wants human sign-off before it leaves draft, not just a green bot.

---
_Generated by [Claude Code](https://claude.ai/code)_

ghost (approved):
## 🔍 Arkana Code Review — follow-up on `9ed13de` (removeCapital guard)

Previous review requested `require(newTargetFiat > 0)` in `removeCapital`. This commit delivers it correctly:

- **`removeCapital` (line ~170):** Guard added after `newTargetFiat` computation, before the collateral ratio check. Correct placement — blocks the negative-minCollateral drain vector.
- **`redeem`/`withdraw`:** Deliberately left without a hard guard. The reasoning in the added comments is sound: these are terminal settlement functions where the clamp (`if (claimRaw <= 0)`) is the correct invariant — it routes all collateral to the long leg, which is the right outcome for a wiped claim. A `require` would block legitimate settlement. This is actually better than what I suggested.
- **Doc fixes:** `fundingRatePerSec` comment now correctly documents the runtime vs. construction asymmetry. Split/merge deferred note added to the design doc.
- **JSON artifact:** Recompiled cleanly — `OP_GREATERTHAN` opcode present in `removeCapital` script, additional `comparison` witness schema entry matches.

All requested changes resolved. Approving.

**⚠️ PROTOCOL-CRITICAL — human sign-off still required.** This is a new settlement contract that handles real BTC collateral with oracle-priced exits.

ghost (approved):
## 🔍 Arkana Code Review — follow-up on `d93a3ef` (oracle hardening + negative tests)

**Diff reviewed:** `9ed13de..d93a3ef` (the only new commit since my last approval).

### Changes

1. **`tests/inventory_hedge_test.rs`** — Three improvements:
   - `test_oracle_paths_verify_price_and_divide`: now asserts `>=2 OP_CAT` and `OP_SHA256` in every oracle-gated function, pinning the on-chain message reconstruction (`sha256(ticker || price || time)`). This mirrors the `stability_vault_test` pattern and closes the gap where a compiler regression could silently drop the concat/hash steps while still passing the old sig-only check. ✅
   - New `test_add_capital_does_no_oracle_call`: negative test ensuring `addCapital` never grows an oracle dependency. Directly addresses finding #5 from my first review. ✅
   - `test_redeem_has_clamp_branches` comment: clarified that the dust guard IS the third `OP_IF`, not a fourth branch. Accurate. ✅

2. **`docs/mm-residual-hedge.md`** — All fenced code blocks now have `text` language tags. Pure markdown-lint fix, no content change. ✅

### Verdict

Tests-only + docs-only diff. CI green (build + format + deploy-preview all pass). No protocol-critical code changes — the contract source (`inventory_hedge.ark`) is untouched.

LGTM. Approving.

tiero:
Resolution on the two deeper findings (high-effort review + the line-219/269 overflow comment). Decision: **document, don't change the logic** — both are inherited verbatim from the audited `stability_vault.ark`, and a rewrite would diverge from it. Pushed in `a1aeef0`.

**1. `claimRaw = newTargetFiat * 1e8 / oraclePrice` overflow (🟡, lines 219/269).** Real but **fail-closed**: the compiler emits `OP_MUL64` + `OP_VERIFY`, so an overflow *aborts* settlement — it never yields a wrong payout. It's a liveness ceiling at ~9.2e10 minor units (≈ $922M). The suggested split-division can't be expressed here: **this language has no `%` operator** (grammar is `+ - * /` only), and a subtraction-based remainder would itself overflow at extreme prices — trading one regime for another while diverging from the reference. Documented the ceiling inline instead.

**2. `updateFunding` bricked under a negative *initial* rate.** `require(delta > 0)` can't pass when the current rate is negative, so repricing can't recover. Reachable only via a negative initial `fundingRatePerSec`, which is unsupported and not constructible-against on-chain (no constructor-level guard in the language). Funds are **not** locked — `redeem`/`withdraw` still settle. Documented as a known limitation; the clean fix is a validating factory (à la `stability_offer.ark`), deferred as a fast-follow.

Test/oracle hardening + the nitpicks are in `d93a3ef`. Contract settlement logic is unchanged and remains byte-for-byte aligned with the StabilityVault reference.

---
_Generated by [Claude Code](https://claude.ai/code)_

tiero:
Response to CodeRabbit's two actionable comments (pushed in `8f498c1`):

**✅ Accepted — `removeCapital` negative `collateralRatioPct` (🟠 Major, outside-diff).** Good catch, and a genuinely new one: with `collateralRatioPct <= -100`, `minCollateral` goes non-positive and the ratio check trivially passes — the long leg could drain a live claim. Added `require(collateralRatioPct >= 0, "invalid collateral ratio")` at the top of `removeCapital`, the same in-function defense-in-depth pattern as the existing `newTargetFiat > 0` guard. (Context: `stability_vault.ark` has the identical formula but seals this at `StabilityOffer.take()` time with `collateralRatioPct >= 100`; `InventoryHedge` ships no factory, so the in-function guard is the equivalent protection.) Artifact recompiled, tests green.

**❌ Declined (maintainer decision) — relax `updateFunding`'s anti-grief guard (`!= 0` → `> 0`).** This exact change was proposed twice and the maintainer explicitly chose **leave as-is, document as unsupported** (see the KNOWN LIMITATION comment added in `a1aeef0`). Rationale: the suggested relaxation is not as harmless as it looks — with `rate > 0` gating, a vault holding a *negative* rate could call `updateFunding` with `delta == 0` accruals (truncation) and silently advance `lastUpdate`, the precise grief the guard exists to block, now on the claim-favoring side. Negative initial rates are unsupported, funds are never locked (`redeem`/`withdraw` settle a wiped claim correctly), and the clean seal is a validating factory (à la `stability_offer.ark`) deferred as a fast-follow. The trade-off is documented inline at the guard.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #42, opened 2026-06-10
Reject identifier shadowing and ABI namespace collisions at validation time

msinkec:
**Summary**

Adds two hard validation errors (in `validate_ast`, before any code generation) closing a class of silent miscompiles. The compiler resolves every variable reference to a name-keyed placeholder (`<name>`) with no scope-qualified lookup, so two distinct bindings that share a name — or two parameters whose names collide *after* ABI expansion — become indistinguishable in the emitted script. Both are now rejected up front.

**What changed**

- **Check 1 — lexical no-shadowing** (`src/validator/mod.rs`): a binding may not reuse a name still live in an enclosing scope. Covers function params shadowing constructor params, locals/loop vars shadowing outer bindings, and `for (x, x)`. Disjoint sibling scopes (separate `if`/`else` branches, separate loops) may reuse names freely. Also rejects assignment to a constructor parameter (they're immutable).
- **Check 2 — expanded-namespace uniqueness** (`src/validator/mod.rs`): parameter names must be unique in the *emitted* placeholder namespace, after array flattening (`xs` → `xs_0..2`), asset decomposition (`foo` → `foo_txid`/`foo_gidx`), and the reserved `serverSig`. Catches collisions invisible at the source level, e.g. `int[] xs` vs `int xs_0`. Reuses the emitter's expansion functions to avoid drift (`collect_lookup_asset_ids`, `decompose_constructor_params` made `pub(crate)` — visibility only, no emission logic changed).
- **Fixture fix** (`tests/bare_vtxo_test.rs`): the legacy contract declared an explicit `signature serverSig` param while the compiler auto-injects the server cosignature — a real duplicate the new check surfaces. Removed the redundant param; ASM output is unchanged.
- **Tests** (`tests/no_shadowing_test.rs`): positive sibling-reuse cases, Check 1/Check 2 negatives, and a sweep asserting every `examples/*.ark` still compiles.

**Notes**

- The conservative `{pk}Sig` reservation was dropped after the example sweep: the emitter deduplicates N-of-N signatures by name (`senderSig` reused for a `sender` pubkey), so reserving it produced false positives.
- The `serverSig` reservation matches current emission. It's pinned to today's transitional architecture; once the tapscript/covenant context split lands it should migrate to tapscript context (noted in the design docs).

## Summary by CodeRabbit

* **New Features**
  * Compiler adds stricter name-scoping checks: rejects parameter/variable shadowing and assignments to constructor parameters.
  * Added validation to ensure emitted placeholder names are unique, preventing collisions.

* **Bug Fixes**
  * Cooperative server mode now injects the server signature instead of requiring it as an explicit parameter.

* **Tests**
  * New and updated tests cover shadowing rules, emitted-name collisions, asset name collisions, and example compilation.

### DIFF
--- a/src/compiler/mod.rs
+++ b/src/compiler/mod.rs
@@ -480,7 +480,7 @@ pub fn compile(source_code: &str) -> Result<ContractJson, String> {
 }
 
 /// Collect all asset ID parameter names used in AssetLookup expressions
-fn collect_lookup_asset_ids(contract: &crate::models::Contract) -> Vec<String> {
+pub(crate) fn collect_lookup_asset_ids(contract: &crate::models::Contract) -> Vec<String> {
     let mut ids = Vec::new();
     for function in &contract.functions {
         for stmt in &function.statements {
@@ -551,7 +551,7 @@ fn collect_asset_ids_from_expression(expr: &Expression, ids: &mut Vec<String>) {
 
 /// Decompose constructor params: bytes32 params used in asset lookups become _txid + _gidx pairs
 /// Array types (e.g., pubkey[]) are flattened to name_0, name_1, name_2, etc.
-fn decompose_constructor_params(
+pub(crate) fn decompose_constructor_params(
     params: &[crate::models::Parameter],
     lookup_asset_ids: &[String],
 ) -> Vec<crate::models::Parameter> {
--- a/src/validator/mod.rs
+++ b/src/validator/mod.rs
@@ -179,6 +179,8 @@ pub fn validate_ast(contract: &Contract) -> Vec<ValidationIssue> {
         }
     }
 
+    check_shadowing(contract, &mut issues);
+
     issues
 }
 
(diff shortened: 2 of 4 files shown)


ghost (approved):
## Arkana Code Review — arkade-os/compiler#42

**Verdict: Approve (with mandatory human sign-off — protocol-critical)**

This PR closes a real class of silent miscompile bugs where two distinct bindings resolve to the same `<name>` placeholder in emitted Bitcoin Script. The design is sound and the implementation is clean. Detailed findings below.

---

### ✅ Correctness

**Check 1 — Lexical no-shadowing** (`src/validator/mod.rs:check_shadowing`)
- All 5 `Statement` variants are exhaustively matched in both `walk_scope` and `check_ctor_assignment`. Rust's exhaustiveness checking will catch future variant additions — no risk of silent gaps. ✓
- Scope stack correctly pushes/pops frames for `IfElse` branches and `ForIn` bodies, allowing sibling reuse while rejecting nesting conflicts. ✓
- `for (x, x)` is caught by the explicit `index_var == value_var` check *before* the scope-shadow check — correct ordering, no false negative. ✓
- Constructor-param immutability (`check_ctor_assignment`) is a sound complement: since shadowing is now rejected, any `VarAssign` targeting a ctor-param name provably refers to the ctor param itself. ✓

**Check 2 — Expanded namespace** (`src/validator/mod.rs:check_expanded_namespace`)
- Reuses `collect_lookup_asset_ids` + `decompose_constructor_params` from the compiler — no logic duplication, no drift. ✓
- Function-param expansion (array flattening only) mirrors `generate_witness_schema` exactly. ✓
- `serverSig` reservation is correctly unconditional per function (both server and exit variants are always emitted, so any param named `serverSig` collides in the server variant). ✓
- Internal functions correctly excluded (`filter(|f| !f.is_internal)`). ✓
- Intentional omission of `{pk}Sig` reservation is correct per the deduplication logic at `compiler/mod.rs ~690-709` (the emitter checks `existing_sig_names` and skips injection if present). ✓

**Visibility change** (`src/compiler/mod.rs`)
- `collect_lookup_asset_ids` and `decompose_constructor_params` promoted from `fn` to `pub(crate) fn`. No external API change — both functions remain crate-internal. No downstream consumer impact. ✓

**Fixture fix** (`tests/bare_vtxo_test.rs`)
- Surfaces a real implicit duplicate: the old test declared `signature serverSig` explicitly while the compiler auto-injects it. Removing the redundant param is correct. ASM assertion (`asm.len() == 9`) unchanged, confirming no behavioral change. ✓

### ✅ Test Coverage

- Positive cases: distinct names, sibling scope reuse, local reassignment. ✓
- Negative cases: function-param shadowing ctor-param, local shadowing function-param, nested local shadowing enclosing local, inner loop var shadowing outer, loop var shadowing ctor-param, `for (x, x)`, ctor-param assignment. ✓
- Namespace collision cases: array element vs scalar, asset decomposition, `serverSig`. ✓
- Full example sweep (`all_examples_still_compile`) — guards against false positives in existing contracts. ✓

### ✅ Cross-repo Impact

- No external crate depends on `arkade-compiler` (verified via Cargo.toml grep across all repos).
- The visibility change is `pub(crate)`, not `pub` — no public API surface change.
- No proto/type/interface changes that could break SDKs or downstream consumers.

### Minor Observations (non-blocking)

1. **Constructor-vs-constructor expansion collisions**: e.g., `contract Foo(int[] xs, int xs_0)` — this IS caught by `record_name` during ctor-expanded iteration (the second `xs_0` fails insertion). Verified correct, but no explicit test case for it. Consider adding one for completeness.

2. **Error message for constructor expansion collisions**: When two ctor params collide after expansion, the error says "parameters in function '{fname}' collide..." — the collision is actually between constructor params, but it's reported per-function since each function's namespace is checked independently. The message is slightly misleading but still actionable. Minor nit.

3. **Multiple errors for `for (x, x)` where `x` is in scope**: You'll get both "must differ" AND "shadows an in-scope binding" (twice). This is fine — over-reporting is preferable to under-reporting in a compiler for money-handling scripts.

### ⚠️ Protocol-Critical Flag

This PR modifies the compiler validation layer for a Bitcoin L2 script compiler. While the code is correct and well-tested, **miscompiles here can cause loss of funds**. Requesting mandatory human review per protocol.

---

LGTM. Clean design, thorough tests, real bug class eliminated.

🤖 Reviewed by Arkana (Claude Opus)

tiero:
I would address coderabbit comments @msinkec

ghost (approved):
**Incremental review** (commit `c9a02d6` — "cache compile")

Test-only cleanup: caches `compile(src)` result in three positive-case assertions (`accepts_distinct_names`, `allows_reassignment_of_local`, `accepts_sibling_scope_reuse`) to avoid redundant double-compilation. Addresses CodeRabbit nitpicks.

No logic changes. Still LGTM.

Previous approval stands. Human sign-off still required per protocol-critical rules.

🤖 Reviewed by Arkana (Claude Opus)

PR #47, opened 2026-06-18
refactor: Compile comparison logic in Bitcoin Script order

### LINKED ISSUE #39: generate_comparison_asm emits reversed operand order in sibling fast-path branches
## Summary

`generate_comparison_asm` in [`src/compiler/mod.rs`](src/compiler/mod.rs) emits comparison operands in the wrong Bitcoin Script order for almost all of its fast-path `match` arms. They produce `<lhs> OP <rhs>` (operator pushed between the operands) instead of the correct `<lhs> <rhs> OP`.

Bitcoin Script is postfix: `OP_EQUAL` / `OP_GREATERTHANOREQUAL` pop the top two stack items, so the operator must come **after** both operands are pushed. The current order would fail at execution (the op runs before the RHS is on the stack).

## Background

This was discovered in PR #29 (LayerZero markers). The `(Property, "==", Literal)` arm was the first branch with a real consumer (`this.activeInputIndex == 0`) and was fixed in `25473b9` (operand order) + `56bf9c1` (the `== true` dummy short-circuit). The remaining sibling branches were intentionally left untouched there to keep the PR scoped — this issue tracks that deferred cleanup.

## Affected branches (all still reversed)

In `generate_comparison_asm` (`src/compiler/mod.rs`, ~lines 1491-1590):

- `(Variable, ">=", Literal)`
- `(Variable, "==", Variable)`
- `(Variable, ">=", Variable)`
- `(Variable, "==", Property)`
- `(Variable, ">=", Property)`
- `(Literal, "==", Variable)`
- `(Literal, ">=", Variable)`
- `(Literal, "==", Literal)`
- `(Literal, ">=", Literal)`
- `(Literal, "==", Property)`
- `(Literal, ">=", Property)`
- `(Property, "==", Variable)`
- `(Property, ">=", Variable)`
- `(Property, ">=", Literal)`
- `(Property, "==", Property)`
- `(Property, ">=", Property)`

Already correct: `(Property, "==", Literal)` (fixed in #29) and the `_ =>` fallback, which delegates to `emit_comparison_asm`.

## Why it's currently dormant

No example contract in the corpus hits these specific Variable/Literal/Property pairings — their comparisons route through dedicated rules (`time_comparison`, `group_property_comparison`, `emit_binary_op_asm`) or the `_ => emit_comparison_asm` fallback, all of which already emit the correct `left, right, op` order. So no shipped contract miscompiles today, but the next contract that hits one of these arms would silently produce a broken script.

## Suggested fix

`emit_comparison_asm` (same file) already emits the correct order, handles 64-bit operands, maps `this.activeInputIndex` / `this.activeBytecode` to their opcodes (via `emit_expression_asm`), and short-circuits the `== true` introspection dummy. The cleanest fix is to **delete the reversed fast-path arms and let everything fall through to `emit_comparison_asm`**, rather than flipping each arm by hand.

Add regression tests asserting `left, right, OP` order for representative pairings (a couple already exist in `tests/packet_primitives_test.rs`).

## Notes

`src/compiler/mod.rs` is a supervised/high-impact file (script-generation semantics) per `CLAUDE.md`, so the change should be reviewed carefully and validated against the full example corpus (regenerate `examples/**/*.json` and diff).

shubertm:
Resolves #39
@tiero

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed comparison behavior for `==` and `>=` so both values are now evaluated consistently before the comparison runs.
  * Improved reliability across comparisons involving variables, literals, and properties.


ghost (changes requested):
## 🔴 Code Review — arkade-os/compiler#47

### Bug: Missed match arm — `(Property, ">=", Property)` still emits reversed order

**`src/compiler/mod.rs` ~line 1478 (PR branch)**

The PR fixes 15 of the 16 affected arms listed in #39 but **misses the last one**. In the PR branch:

```rust
(Expression::Property(prop), ">=", Expression::Property(prop2)) => {
    asm.push(format!("<{}>", prop));
    asm.push(OP_GREATERTHANOREQUAL.to_string()); // ❌ opcode BEFORE second operand
    asm.push(format!("<{}>", prop2));
}
```

Should be:

```rust
(Expression::Property(prop), ">=", Expression::Property(prop2)) => {
    asm.push(format!("<{}>", prop));
    asm.push(format!("<{}>", prop2));             // ✅ second operand first
    asm.push(OP_GREATERTHANOREQUAL.to_string()); // ✅ opcode last
}
```

Bitcoin Script is postfix. `OP_GREATERTHANOREQUAL` pops the top two stack items. If the opcode fires before the RHS is pushed, the script is malformed. This would affect any contract comparing two introspection properties with `>=` (e.g., `tx.weight >= tx.locktime`).

---

### Missing test coverage

The PR adds 11 tests but there are 16+ comparison arms. Missing tests for:

- `Variable == Literal` (code fixed, no test)
- `Variable == Property` (code fixed, no test)
- `Variable >= Property` (code fixed, no test)
- `Literal == Property` (code fixed, no test)
- `Literal >= Property` (code fixed, no test)
- **`Property == Property`** (code fixed, no test)
- **`Property >= Property`** (code **NOT fixed**, no test — this is the bug above)

The missing `Property >= Property` test is particularly concerning — if it existed, it would have caught the bug.

---

### Architectural note (non-blocking)

Issue #39 itself recommended deleting all fast-path arms and letting everything fall through to `emit_comparison_asm`, which already emits the correct `left, right, op` order and handles 64-bit operands. The manual arm-flipping approach taken here is error-prone (as the missed arm proves). Consider the delete-and-fallthrough approach for a more robust fix — fewer lines to maintain, one source of truth for operand ordering.

---

### Summary

| Finding | Severity | Status |
|---------|----------|--------|
| `(Property, ">=", Property)` still reversed | 🔴 **Bug** — would produce broken Script | Must fix |
| 7 of 16 arms have no regression test | 🟡 Medium | Should fix |
| Manual flipping vs. delete-and-fallthrough | 🟢 Suggestion | Nice to have |

**Requesting changes** for the missed arm. The rest of the fix is correct and the new tests are solid. One-line fix + one more test and this is good to go.

⚠️ **Protocol-critical**: This function generates Bitcoin Script. Requires human sign-off before merge even after the fix.

ghost (approved):
## ✅ Re-review — arkade-os/compiler#47 (commit `6184c19c`)

### Previous finding: FIXED ✅

The `(Property, ">=", Property)` arm at `src/compiler/mod.rs:1476` now correctly emits `lhs rhs OP_GREATERTHANOREQUAL` in postfix order. All 16 fast-path comparison arms are verified correct.

### Remaining note: test coverage gaps (non-blocking)

7 of 16 modified arms still lack dedicated tests:

| Arm | Operator |
|---|---|
| `Variable == Literal` | `==` |
| `Variable == Property` | `==` |
| `Variable >= Property` | `>=` |
| `Literal == Property` | `==` |
| `Literal >= Property` | `>=` |
| `Property == Property` | `==` |
| `Property >= Property` | `>=` |

The 11 existing tests provide good coverage of the pattern — every expression type appears as LHS, and both operators are exercised. The untested arms follow the exact same mechanical structure, so regression risk is low. I'm not blocking on this, but recommend adding them as a follow-up to make the regression suite complete.

LGTM. Approving.

msinkec:
Removed comaprison tests, since we're still missing symbolic stack management, which will probably require a lot of rewriting of this. The test would only churn in that case.

msinkec (approved):

PR #75, opened 2026-08-03
Static arrays with declared sizes

msinkec:
Arrays were fixed at three elements by a single compile-wide constant (`DEFAULT_ARRAY_LENGTH`). This replaces that with sCrypt-style static arrays: the size is part of the type and every layer reads it from the declaration.

```
contract Quorum(pubkey[5] oracles) {
  function attest(signature[5] sigs, bytes32 msg, int index) {
    int[3] weights = [1, 2, 5];
    weights[0] = 4;

    for (i, sig) in sigs { ... }          // unrolls 5x
    require(checkSigFromStack(sigs[index], oracles[index], msg));  // bound-checked against 5
    require(total >= oracles.length);     // folds to 5
  }
}
```

## What changed

- **`T[N]` is the only array form.** Bare `T[]` no longer parses; `T[0]` and `T[01]` are parse errors. `DEFAULT_ARRAY_LENGTH` is deleted.
- **`arr.length`** folds to the declared size at compile time.
- **Local arrays**: `int[3] xs = [1, 2, 3];` with element assignment at a literal index. Elements bind exactly like parameter arrays, so indexing, loops and `.length` treat them identically.
- **Artifact ABI groups arrays.** `constructorInputs`, `arkade.inputs` and leaf `witness` keep one entry per source parameter, carrying the size: `{ "name": "oracles", "type": "pubkey[3]" }`.
- **Arrays are covenant-only.** Constructor and covenant function parameters take arrays; `tapscript` function inputs must be scalars, so no tapleaf witness carries an array for now.
- **`for … in tx.assetGroups` is rejected.** The group count is not known at compile time, so the old fixed unroll silently checked only the first three groups.

Rejected with a message: bare `T[]`, zero or leading-zero sizes, an array input on a `tapscript` function, a literal index at or past the declared size, an initializer whose element count or type does not match, an array literal without a declared array type, and `xs[i] = v` at a runtime index.

## Notes for review

**The compiler derives array lengths from the symbolic stack.** `Generator::array_length` counts the contiguous `$array:name:i` bindings already tracked there, so no new state was needed for the bound check or the unroll count, and local arrays work through the same path as parameters.

**The ABI and the script now speak two representations.** The artifact groups arrays; `asm` keeps per-element placeholders (`<oracles_0>`). `validate_output` derives the expected constructor prologue by flattening `constructorInputs`, so a mismatch between the two fails the build. Consumers expand: `arkade-bindgen` (`fields_from_ark_type`) and `playground/codegen.js` (`expandFields`). Without that expansion, `Encoding::from_ark_type("pubkey[3]")` would have fallen through to `Unknown` and generated `[]byte` fields for a pubkey — silently wrong stubs, hence the new `test_ir_expands_grouped_array_fields`.

**Runtime-index assignment (`xs[i] = v`) is deferred, not forgotten.** The VM has `OP_PICK` (read at a runtime depth) but no inverse, so every write is emulated — `Generator::assign` costs `O(depth)` opcodes today for *every* assignment. Emulating a dynamic array write on the current opcode set costs `N` guarded writes. The design doc writes up an `OP_PUT` proposal (dual of `OP_PICK`, O(1) interpreter cost, constant `-2` stack effect) that would make it ~11 opcodes and collapse `assign` to two. Note the `0 <= i < N` check stays the compiler's job either way: `OP_PUT` can only bound-check against the whole stack, while an array occupies a subrange, so a witness-supplied index could otherwise overwrite a neighbouring binding.

## Testing

`cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check` pass at each of the six commits. `./playground/build.sh` succeeds.

New e2e (`sh scripts/e2e.sh`): `tests/e2e/contracts/static_arrays.ark` runs an `int[4]` constructor array, an `int[5]` function array and a local `int[3]` through the Arkade VM, with runtime indexing into both a parameter and a local array, and rejection cases for negative, past-the-end and local out-of-range indices. The harness builds the witness from the ABI (`covenantWitness`/`expandInput`), so it exercises the documented client-side expansion rather than assuming an order.

Two things worth knowing about that e2e, from mutation-checking it:

- Reversing the local-array push order initially *passed*, because flipping the order flips the binding names too and literal indexing stays correct by name. Only a runtime index into a local array depends on that layout, so the contract now does one; the mutation fails the suite.
- An off-by-one in the runtime bound value still passes the e2e — an over-long guard lets the read succeed and the later total check trips the same `OP_VERIFY failed`. That case is pinned by the unit test `runtime_index_bound_check_uses_the_arrays_own_size`, which asserts the exact emitted length.

## Migration

Every `T[]` declaration needs a size; `[3]` preserves current behaviour. Clients expand a grouped array entry into `name_0 … name_{N-1}` in index order (witness order is reverse declaration order, element 0 closest to the top).

## Summary by CodeRabbit

* **New Features**
  * Added fixed-size arrays with indexing, `.length`, literals, assignments, and loop iteration.
  * Preserved grouped array inputs in contract interfaces while expanding elements for execution.
  * Added validation for array sizes, bounds, and invalid array usage.
  * Added end-to-end coverage for static-array contracts and transaction scenarios.

* **Documentation**
  * Documented array behavior, stack ordering, serialization, and constructor handling.
  * Updated the threshold oracle example to use fixed-size arrays.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ contract FujiSafe(
 - `bool`: Boolean value
 - `asset`: Asset identifier (for asset-aware contracts)
 
-Any type can be declared as a fixed-size array by appending a size: `pubkey[5] oracles`, `signature[5] sigs`. The size is part of the type and must be a positive integer literal; there is no unsized array type. Array parameters are flattened into one input per element (`oracles_0 … oracles_4`), `for` loops over them unroll once per element, and each element is one stack item — so large sizes grow the compiled script proportionally.
+Any type can be declared as a fixed-size array by appending a size: `pubkey[5] oracles`, `signature[5] sigs`. The size is part of the type and must be a positive integer literal; there is no unsized array type. Array parameters are flattened into one input per element (`oracles_0 … oracles_4`), `for` loops over them unroll once per element, and each element is one stack item — so large sizes grow the compiled script proportionally. Arrays are allowed in constructor and covenant function parameters; `tapscript` function inputs must be scalars.
 
 ### Contract Structure
 
--- a/README.md
+++ b/README.md
@@ -293,6 +293,8 @@ contract FujiSafe(
 - `bool`: Boolean value
 - `asset`: Asset identifier (for asset-aware contracts)
 
+Any type can be declared as a fixed-size array by appending a size: `pubkey[5] oracles`, `signature[5] sigs`. The size is part of the type and must be a positive integer literal; there is no unsized array type. Array parameters are flattened into one input per element (`oracles_0 … oracles_4`), `for` loops over them unroll once per element, and each element is one stack item — so large sizes grow the compiled script proportionally.
+
 ### Contract Structure
 
 An Arkade Language file may start with zero or more `import` declarations, followed by a `contract` declaration:
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ This language significantly lowers the barrier for Bitcoin-native app developmen
 
 ## Emulator E2E Tests
 
-The self-contained E2E suite builds the compiler and runs the counter and HTLC artifacts through the Arkade VM and btcd tapscript interpreter.
+The self-contained E2E suite builds the compiler and runs the counter, HTLC, symbolic-stack and static-array artifacts through the Arkade VM and btcd tapscript interpreter.
 
 ```bash
 ./scripts/e2e.sh -v
--- a/README.md
+++ b/README.md
@@ -430,21 +430,23 @@ Arkade Language compiles to Arkade Script and produces a JSON artifact for use w
 | Field                 | Description                                                                               |
 |-----------------------|-------------------------------------------------------------------------------------------|
 | `contractName`        | Contract identifier                                                                       |
-| `constructorInputs`   | Constructor parameters in source declaration order after array expansion                  |
+| `constructorInputs`   | Constructor parameters in source declaration order, one entry per parameter               |
 | `functions`           | Spend groups: `{ name, arkade?, leaves[] }`                                               |
 | `arkade`              | Optional emulator-run covenant `{ inputs, asm }`                                          |
-| `arkade.inputs`       | Function parameters in source declaration order after array expansion                     |
+| `arkade.inputs`       | Function parameters in source declaration order, one entry per parameter                  |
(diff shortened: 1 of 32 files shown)


msinkec:
Addressed the CodeRabbit review. Three of the four findings reproduced and are fixed in 787d46e; one no longer applies.

**Fixed — `ArrayLiteral` missing from two tree-walking passes.** Both confirmed by repro:

- `substitute_expression` (`src/compiler/loops.rs`): a local array declared in a loop body with loop variables as elements failed to compile with `undefined binding 'i'`.
- `rewrite_expression_concat` (`src/compiler/concat.rs`): `bytes[2] parts = [a + b, a]` on two `bytes32` operands emitted `OP_ADD` instead of `OP_CAT` — a silent miscompile, the most severe of the batch.

Both now recurse into elements, with regression tests asserting no `$array:` binding leaks from the loop case and that the concat case emits `OP_CAT` and no `OP_ADD`.

**Fixed — heterogeneous array literals.** `bytes32[2] xs = [h, 5]` compiled clean, because the inferred array type only carries the first element's type. Each element is now checked against the declared element type.

One correction to the suggested fix: it proposed rejecting elements "compatible but not equal" to the declared type. Implemented as strict equality first, that rejects `bytes[2] parts = [a + b, a]`, since `bytes32` in a `bytes` slot is a widening the language allows everywhere else. The check now uses the same `binding_types_compatible` rule as the binding-level check, so `bytes32` into `bytes` passes and `int` into `bytes32` is rejected.

**Fixed — assignment at a non-zero literal index.** Test added. Worth noting the concern was already covered end to end (`scale[1] = 10` in the e2e contract, verified through the VM), but the unit test gives faster feedback and pins the exact depth: index 0 rolls at depth 1 with no walk-back, index 1 rolls at depth 2 and swaps back.

**Not applicable — per-element `injected` status on grouped witness arrays.** 0784588 rejects array-typed inputs on `tapscript` functions outright, so no leaf witness can carry an array and a mixed signature array is unrepresentable. Worth revisiting if array witnesses are enabled for tapscripts later.

`cargo test --workspace`, clippy `-D warnings`, `cargo fmt --check` and the full e2e suite pass.

PR #89, opened 2026-09-09
Compare whole arrays and structs with == and !=

msinkec:
`require(arr0 == arr1)` and `require(point0 == point1)` now compile instead of being rejected as composite values. Nested structs, struct fields of struct or array type, array literals, private-call results, and the native result structs (`Outpoint`, `AssetId`, `ECPoint`) all work on either side.

## How it compiles

Both operands are laid out on the stack through the existing `emit_typed_value` (first leaf on top), then each pair is rolled together:

- `==` → `<n> OP_ROLL OP_EQUALVERIFY` per pair; the last pair needs no roll since `OP_EQUALVERIFY` is symmetric
- `!=` → pair results are folded with `OP_BOOLAND`, then `OP_NOT OP_VERIFY`

```
require(keys == other);
<keys.1> <keys.0> OP_1 OP_PICK OP_1 OP_PICK OP_5 OP_PICK OP_5 OP_PICK OP_2 OP_ROLL OP_EQUALVERIFY OP_EQUALVERIFY
```

## Changes

- `src/compiler/comparison.rs` — `emit_composite_requirement`, plus `composite_type` and a `roll` helper on `Generator`
- `src/compiler/mod.rs` — `Generator` carries a typechecker `Scope` built from the existing `build_scope_with_structs`, so it knows an operand's declared type and leaf count; `Requirement::Comparison` routes composites to the new emitter
- `src/compiler/functions.rs` — private-function parameters register their types for the callee body
- `src/typechecker/mod.rs`, `src/validator/mod.rs` — composite operands are allowed in a comparison; `>`/`<` and friends stay rejected, and mismatched layouts are a hard error (`comparison '==' is not defined between 'int[3]' and 'int[2]'`). Composites in any other value position stay rejected as before
- `README.md` — the "whole-struct comparison is not supported" line documents the feature instead

## Tests

Three tests that asserted the old rejection became positive tests (array `==`/`!=`, nested-struct `==` plus layout mismatch, whole-outpoint `==`).

The Go e2e suite executes composite comparisons on the real VM: struct `==`, struct `!=`, array-vs-literal `==`, native `Outpoint`/`AssetId`/`ECPoint` `==`, and `original == result(value)` against a private-call result. Each new check was confirmed to bite — flipping an expected element fails with `OP_EQUALVERIFY failed`, and `supplied != supplied` fails with `OP_VERIFY failed`.

`cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check`, and `./scripts/e2e.sh` all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019uB2qKYUVTjJRPDCToNxmJ

## Summary by CodeRabbit

* **New Features**
  * Added support for comparing complete arrays and structs with `==` and `!=`.
  * Comparisons evaluate each element or field, including nested composites and built-in composite values.
  * Composite comparisons require matching declared types; other comparison operators remain unsupported.
  * Added support for multi-file imports, recursive dependency loading, and imported helpers.

* **Documentation**
  * Documented compiler and WebAssembly APIs, source bundles, constructor validation, and file-aware warnings.
  * Clarified function, constant, indexing, array, and struct comparison behavior.

* **Tests**
  * Added coverage for imports, bundles, and composite comparisons.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,7 +252,7 @@ Comments use `//`. Identifiers start with a letter and contain letters, digits, 
 | `struct` | User-declared, nested structs and scalar arrays allowed |
 | `AssetId`, `Outpoint`, `ECPoint` | Native result structs: `{txid, gidx}`, `{txid, vout}`, `{x, y}` |
 
-Arrays and structs are allowed in constructor and covenant parameters and as locals. Tapscript inputs must be scalars. Arrays of structs, whole-struct assignment, and whole-struct comparison are not supported.
+Arrays and structs are allowed in constructor and covenant parameters and as locals. Tapscript inputs must be scalars. Arrays of structs and whole-struct assignment are not supported. Whole arrays and structs compare with `==` and `!=` inside `require`, leaf by leaf, and both sides must have the same declared type.
 
 ### Functions
 
--- a/src/compiler/comparison.rs
+++ b/src/compiler/comparison.rs
@@ -1,5 +1,7 @@
 use super::*;
 use crate::models::*;
+use crate::opcodes::OP_ROLL;
+use crate::typechecker::{bind_local_type, infer_type, ArkType};
 
 /// Emit assembly for a comparison requirement.
 ///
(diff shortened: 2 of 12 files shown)


msinkec:
Thanks both — pushed `e42a16b`. Summary of what I took and what I skipped.

## Fixed

**Scope leaking past a private call** (coderabbit `functions.rs:183`, arkana Issue 1) — valid, and it reproduced:

```
struct Res { int selected; int other; }
private function same(Res left, Res right) { require(left == right); }
function spend(Res a, Res b, int left) { same(a, b); require(left == 1); }
// before: undefined binding 'left.other'
```

While confirming it I found the same class of leak at **block** boundaries, which neither review caught — a struct local in an `if` body kept its type after `exit_scope` dropped its stack bindings:

```
if (flag == 1) { Point q = { x: 1, y: 2 }; require(q.x == 1); }
let q = 5;
require(q == 5);   // before: undefined binding 'q.y'
```

So rather than only mirroring `caller`/`self.stack` in `emit_call`, the type scope now follows the same lifetime discipline as the stack everywhere: `enter_scope` snapshots it, `exit_scope` restores it, and `emit_call` restores it alongside `self.stack`. Both cases are covered by `composite_types_do_not_outlive_their_bindings`.

**Constructor vs function parameter priority** (arkana Issue 2) — flipped, function parameters now win. Beyond matching the typechecker, it matches the physical stack: `binding_index` scans from the bottom, where function inputs sit, so function parameters already won there.

**Exact composite type equality** (coderabbit `validator/mod.rs:1379`, arkana "new finding") — taken, and it deletes code. One correction to the reasoning though: type errors are *not* a gate — `check_contract` results become non-fatal `warnings` on `ContractJson`, so the validator is the only gate and `bytes[1] == bytes20[1]` did compile. It's now rejected there, matching `check_comparison` and the README rule. Regression test added.

## Skipped

**Validate every array-literal element** (coderabbit `validator/mod.rs:1391`) — declining. `[1, true]` is not a miscompile: `true` emits `OP_1`, so it compiles to exactly the same script as `[1, 1]`. For genuinely mixed elements (`int[2] == [1, someBytes32]`) the equivalent scalar comparison `nums[1] == someBytes32` is only a *warning* in this language, so a hard validator error here would make composites stricter than their scalar counterpart. If the diagnostic is wanted, the right place is `infer_type`'s `ArrayLiteral` branch (which today takes the first element's type) so scalars and composites report it the same way — that's a separate change.

**`width == 0` guard** — declining, it is unreachable: `struct Empty { }` fails with `expected identifier` and `int[0]` with `expected array_size`, so no zero-width type can be constructed. I removed the guard deliberately in a YAGNI pass. If the grammar ever admits an empty struct, the guard becomes necessary — happy to add it then, or now if you'd rather have the belt.

`cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check` and `./scripts/e2e.sh` all pass.

msinkec:
Latest arkana pass (`ff829da`) closes every substantive item and leaves two nits, both asking for explanatory comments — one on `composite_type` (that `Unknown` falls through to the scalar path) and one on the routing in `generate_requirement_asm` (that mismatched composite types are already rejected by the validator). Declining both: comments in this repo describe what the code *is*, not why a reviewer had to reason about it, and neither is a simplification with a known ceiling. arkana marks them non-blocking itself.

What I did do this pass:

**Merged `master` (`b01ff18`).** The PR had gone `CONFLICTING` after #90 (contract imports + README rewrite) landed. The only conflict was the arrays/structs paragraph, which #90 rewrote; resolved in master's new voice:

> Read and assign fields individually; `require` compares whole arrays and structs with `==` and `!=` when both sides have the same declared type.

**Re-verified the head independently**, since `ff829da` landed without me:

- Heterogeneous literals are now caught with useful messages — `int[2] == [1, true]`, `[1, true] == [1, 2]`, and `Point == { x: 1, y: <bytes32> }` all fail as `expected 'int', got 'bool'` / `'bytes32'`; `[1, 2] == [1, 2, 3]` reports both the element-count and the type mismatch. Valid literals (`nums == [1, 2, 3]`, `p == { x: 1, y: 2 }`, `keys == [k0, k1]`) still compile.
- No new miscompile paths: whole-array assignment is still rejected, and the grammar rejects a composite `==` nested inside another expression, so the composite lowering stays reachable only from `require`.
- Composite comparisons work across the new import feature — an imported `Point` compares as expected and `new Inner(origin)` still flattens to `<VTXO:Inner(<origin.x>,<origin.y>)>`.
- Full suite green on the merge: 310 feature tests, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check`, `./scripts/e2e.sh`.

One heads-up unrelated to the code: `scripts/pre-commit` breaks merge commits. Its `git stash --keep-index --include-untracked` clears `MERGE_HEAD`, so `git commit` then dies with `fatal: could not open '.git/MERGE_HEAD'`. I committed the merge with `--no-verify` after running the same checks by hand. Worth guarding the hook with `[ -e .git/MERGE_HEAD ] && exit 0`.

`reviewDecision` is still `CHANGES_REQUESTED` from arkana's 2026-09-09 review; its two later passes both say ready to merge, so that just needs the stale review dismissed.

